PR #26, opened 2016-08-11
Reconnect

ebuchman:
This is the tmsp part of https://github.com/tendermint/tendermint/issues/243

- use `OnReset` from https://github.com/tendermint/go-common/pull/2 to allow tmsp client to be restarted
- StopForError will Stop, Reset, and Start, dropping back into the connect loop. 
- Reset will flush the queue, which fixes https://github.com/tendermint/tmsp/issues/9
- Consumers can register a callback upon successful connection (eg. to start a service) with `SetConnectCallback`. In order to make this work, `New` no longer runs `Start`, as its now the consumer's responsibility to do so, presumably after setting the callback.


ebuchman:
replaced by https://github.com/tendermint/tmsp/pull/27

PR #31, opened 2016-09-12
Make console more robust to typos

ethanfrey:
If I make a typeo in the console, it will die.  Both forgetting to pass an argument to `query` as well as a typo `check` not `check_tx`.  This is not user-friendly.

As I am playing with tmsp-cli as a way to debug and understand apps, I added error messages to replace exit cases.

I'd also like to add readline support (like auto-complete, up arrow to go through buffer history), but that is another project.  Let me know if you would like that.

### DIFF
--- a/cmd/tmsp-cli/tmsp-cli.go
+++ b/cmd/tmsp-cli/tmsp-cli.go
@@ -123,6 +123,14 @@ func before(c *cli.Context) error {
 	return nil
 }
 
+// badCmd is called when we invoke with an invalid first argument (just for console for now)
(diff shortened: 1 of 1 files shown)


ebuchman:
Thanks for this! I've been annoyed by this too but never did anything about it. As for readline support, just use `rlwrap`!  I think that's much preferable than implementing all the functionality into this little tool.

ethanfrey:
Nice. I never heard of rlwrap before but it looks super useful. Now time to
get coding...

PR #41, opened 2016-12-24
Updates, Closes #40, Closes #35

### LINKED ISSUE #32: Make abci-cli output more machine parseable
either pack it all in json or just print the code and the data

### LINKED ISSUE #35: dont exit console on error
https://github.com/urfave/cli/issues/565

### LINKED ISSUE #40: Persistent flags for console?
Currently, flags applied to a console do not get automatically become applied to console commands. Does it make sense to implement this for some or all flags (--verbose maybe)? Interestingly, you can still type the flags in the beginning of the console command (the command `--verbose echo hello` works within console)

rigelrozanski:
Fixed flag persistence when using console or batch, no longer exits on error from console or batch. Also added query support for Counter example app (ebuchman request, not an issue)

### DIFF
--- a/cmd/tmsp-cli/tmsp-cli.go
+++ b/cmd/tmsp-cli/tmsp-cli.go
@@ -135,6 +135,28 @@ func badCmd(c *cli.Context, cmd string) {
 	cli.DefaultAppComplete(c)
 }
 
+//Generates new Args array based off of original Args call to maintain flag persistence
+func persistentArgs(removeArg string, line []byte) []string {
+	//function to remove first slice of matching string from string array
+	remove := func(slice []string, removeText string) []string {
+		for i := len(slice) - 1; i >= 0; i-- { //search from end to start
+			if slice[i] == removeText {
+				return append(slice[:i], slice[i+1:]...)
+			}
+		}
+		return slice
+	}
+
+	//generate the arguments to run from orginal os.Args
+	// to maintain flag arguments
+	args := os.Args
+	args = remove(args, removeArg)
+	if len(line) > 0 { //prevents introduction of extra space leading to argument parse errors
+		args = append(args, strings.Split(string(line), " ")...)
+	}
+	return args
+}
+
 //--------------------------------------------------------------------------------
 
 func cmdBatch(app *cli.App, c *cli.Context) error {
@@ -150,12 +172,9 @@ func cmdBatch(app *cli.App, c *cli.Context) error {
 		} else if err != nil {
 			return err
 		}
-		args := []string{"tmsp-cli"}
-		if c.GlobalBool("verbose") {
-			args = append(args, "--verbose")
-		}
-		args = append(args, strings.Split(string(line), " ")...)
-		app.Run(args)
+
+		args := persistentArgs("batch", line)
+		app.Run(args) //cli already prints error within its func call
 	}
 	return nil
 }
@@ -163,6 +182,7 @@ func cmdBatch(app *cli.App, c *cli.Context) error {
 func cmdConsole(app *cli.App, c *cli.Context) error {
 	// don't hard exit on mistyped commands (eg. check vs check_tx)
 	app.CommandNotFound = badCmd
+
 	for {
 		fmt.Printf("\n> ")
 		bufReader := bufio.NewReader(os.Stdin)
@@ -173,8 +193,7 @@ func cmdConsole(app *cli.App, c *cli.Context) error {
 			return err
 		}
 
-		args := []string{"tmsp-cli"}
-		args = append(args, strings.Split(string(line), " ")...)
+		args := persistentArgs("console", line)
 		app.Run(args) //cli already prints error within its func call
 	}
 }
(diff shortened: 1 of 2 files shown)


cmd/tmsp-cli/tmsp-cli.go
```diff
@@ -38,6 +76,10 @@ func main() {
 			Name:  "verbose",
 			Usage: "print the command and results as if it were a console session",
 		},
+		cli.BoolFlag{
+			Name:  "mro",
```
ebuchman: lets just call this `json`. Not sure `mro` is a thing :P

cmd/tmsp-cli/tmsp-cli.go
```diff
@@ -15,10 +16,47 @@ import (
 	"github.com/urfave/cli"
 )
 
+//structure for data passed to print response
+// variables must be exposed for JSON to read
+type pr struct {
```
akalyaev: not sure this is a good naming

p.s. here is a nice presentation from google dev https://talks.golang.org/2014/names.slide#1

rigelrozanski: Cool, thanks for the feedback/presentation! Any thoughts as to a better struct name? the current `pr` just comes from the fact it's fed to printResponse. Maybe `prVar`, `prOut`, or maybe something more to do with results, as it's essentially just a more detailed output of its variable `Res types.Results`. So with this in mind maybe... `expRes` or `expRslt` for expanded results

ebuchman: How about just `Response`?

rigelrozanski: I like it, changing now

cmd/tmsp-cli/tmsp-cli.go
```diff
@@ -15,10 +16,47 @@ import (
...
+	Res       types.Result
+	S         string
```
akalyaev: What is `S`?

cmd/tmsp-cli/tmsp-cli.go
```diff
@@ -146,19 +207,22 @@ func cmdBatch(app *cli.App, c *cli.Context) error {
 		} else if err != nil {
 			return err
 		}
-		args := []string{"tmsp-cli"}
-		if c.GlobalBool("verbose") {
-			args = append(args, "--verbose")
+
+		args := persistentArgs(line)
+		err2 := app.Run(args) //cli prints error within its func call
```
akalyaev: Is it considered a bad style to reuse `err` in go?

cmd/tmsp-cli/tmsp-cli.go
rigelrozanski: was just about to comment on this, This is a relic from the old code. Holds information typically a variant of `Res.Data` in a string (but sometimes slightly different). I've just changed `S` to `Data`

cmd/tmsp-cli/tmsp-cli.go
rigelrozanski: I prefer to distinguish the name to make the code less ambiguous. It can also help simplify debugging for future code. I'd opt to maintain use of `err2`

cmd/tmsp-cli/tmsp-cli.go
rigelrozanski: addressed in commit 'naming changes'

example/counter/counter.go
```diff
@@ -72,5 +72,14 @@ func (app *CounterApplication) Commit() types.Result {
 }
 
 func (app *CounterApplication) Query(query []byte) types.Result {
-	return types.NewResultOK(nil, Fmt("Query is not supported"))
+	queryStr := string(query[:])
```
ebuchman: shouldn't need the `[:]` since its already a byte slice.

example/counter/counter.go
```diff
@@ -72,5 +72,14 @@ func (app *CounterApplication) Commit() types.Result {
...
+
+	switch queryStr {
```
ebuchman: let's just use `hash` and `tx`, no need for `Count`. Also, let's just  return the value, no need for the extra english.

Finally, seems like `ErrUnknownRequest` might be more appropriate

ebuchman:
Can we split this out into two PRs? 
One with the exit fix and the counter query, the other with the mro.

I think the mro stuff needs more thought, json doesnt seem like a good idea to me anymore (sorry my bad). The point of making the output machine readable is to simplify some of the tests or to make it easier to write more.

cmd/tmsp-cli/tmsp-cli.go
rigelrozanski: Oki doki - will remove

rigelrozanski:
Yes, I will remove MRO stuff from this PR, and will open a new one once we come to a better idea on what specifically we want.. maybe continue MRO discussion in issue #32

example/counter/counter.go
rigelrozanski: Okay will need to add hash and tx to struct, not currently stored

ebuchman: hmm? I just mean instead of the query being `hashCount` it can just be `hash`

rigelrozanski: aha, thanks for clarifying

ebuchman:
Rebased in https://github.com/tendermint/tmsp/pull/46

PR #53, opened 2017-01-17
Makefile: add fmt and lint targets

roylee17:
and run `make fmt`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,33 @@
-.PHONY: all test get_deps
+.PHONY: all test fmt lint get_deps
 
 all: protoc install test
 
 NOVENDOR = go list github.com/tendermint/abci/... | grep -v /vendor/
 
 protoc:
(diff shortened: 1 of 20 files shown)


melekes:
Thanks for this! Not sure we need `fmt` and `lint` targets, though. https://github.com/fatih/vim-go, which I am using, does this by default every time I save .go file. For PRs, we could enable https://houndci.com/ or similar bot and write some `CONTRIBUTING.md` guidelines (https://github.com/blog/1184-contributing-guidelines). What do you think?

roylee17:
For Atom, and VS code, the lint/vet/fmt deal with the current open tab, and not the whole codebase. It's convenient if we can scan whole codebase and fix similar ones at a time instead of file by file.

Maybe we can revert it once the lint errors converges, and we can strictly enforce CI to reject warnings.

PR #68, opened 2017-04-05
Comment out MessageType

KrzysiekJ:
As it is an unused enum, leaving it uncommented causes dead Protocol Buffers code to be generated.


melekes:
Maybe it should be removed?

adrianbrink:
@ebuchman Can we safely remove it?

ebuchman:
Yes I believe we can remove it.

adrianbrink:
Moved to #81

PR #78, opened 2017-05-01
New logging

melekes:


cmd/dummy/main.go
```diff
@@ -17,19 +18,24 @@ func main() {
...
 
+	logger := log.NewTmLogger(os.Stdout)
+
 	// Create the application - in memory or persisted to disk
 	var app types.Application
 	if *persistencePtr == "" {
 		app = dummy.NewDummyApplication()
 	} else {
 		app = dummy.NewPersistentDummyApplication(*persistencePtr)
+		app.(*dummy.PersistentDummyApplication).SetLogger(log.With(logger, "module", "dummy"))
```
melekes: it looks a bit ugly, but don't think that `SetLogger` should be a part of the `Application` interface.

cmd/dummy/main.go
```diff
@@ -17,19 +18,24 @@ func main() {
...
 	}
 
 	// Start the listener
 	srv, err := server.NewServer(*addrPtr, *abciPtr, app)
 	if err != nil {
-		log.Fatal(err.Error())
+		logger.Error(err.Error())
```
melekes: What do you think? Do we need a helper method:

```
func Fatal(logger Logger, msg string, keyvals ...interface{}) {
  logger.Error(msg, keyvals...)
  os.Exit(1)
}
```

My opinion is that one line does not worth it.

example/dummy/persistent_dummy.go
```diff
@@ -38,14 +42,19 @@ func NewPersistentDummyApplication(dbDir string) *PersistentDummyApplication {
 	stateTree := iavl.NewIAVLTree(0, db)
 	stateTree.Load(lastBlock.AppHash)
 
-	log.Notice("Loaded state", "block", lastBlock.Height, "root", stateTree.Hash())
+	// log.Notice("Loaded state", "block", lastBlock.Height, "root", stateTree.Hash())
```
melekes: hm... what should we do here? pass a logger as an argument?

example/dummy/persistent_dummy.go
```diff
@@ -134,8 +146,7 @@ func LoadLastBlock(db dbm.DB) (lastBlock LastBlockInfo) {
 		r, n, err := bytes.NewReader(buf), new(int), new(error)
 		wire.ReadBinaryPtr(&lastBlock, r, 0, n, err)
 		if *err != nil {
-			// DATA HAS BEEN CORRUPTED OR THE SPEC HAS CHANGED
-			log.Crit(cmn.Fmt("Data has been corrupted or its spec has changed: %v\n", *err))
+			cmn.PanicCrisis(errors.Wrap(*err, "cannot load last block (data has been corrupted or its spec has changed)"))
```
melekes: `log.Crit` does not interrupt the program. Can we continue working if the "data has been corrupted"? If yes, I will remove the Panic.

melekes: Note that during `SaveLastBlock` we panic

cmd/dummy/main.go
```diff
@@ -17,19 +18,24 @@ func main() {
...
+		os.Exit(1)
```
melekes: What do you think? Do we need a helper method:

```
func Fatal(logger Logger, msg string, keyvals ...interface{}) {
  logger.Error(msg, keyvals...)
  os.Exit(1)
}
```

My opinion is that one line does not worth it.

melekes:
- `BaseService` now exposing `Logger` field.
-  `Service` interface has a `SetLogger` method

cmd/dummy/main.go
ebuchman: we could do:

```
dummyApp := dummy.New...
dummyApp.SetLogger...
app = dummyApp
```

But thats not much better.

cmd/dummy/main.go
ebuchman: Yeh let's not. At least yet.

example/dummy/persistent_dummy.go
ebuchman: Actually thats not a bad idea. Would avoid the ugly SetLogger with the type assertion too

example/dummy/persistent_dummy.go
ebuchman: Don't think so.

example/dummy/persistent_dummy.go
ebuchman: Fair point. It's not a critical log message to have. I've also considered moving the tree loading outside this contructor, so it would take the tree as an arg. But then we still have the ugly SetLogger issue (which isnt really a big deal ...)

client/socket_client.go
```diff
@@ -53,7 +53,8 @@ func NewSocketClient(addr string, mustConnect bool) (*socketClient, error) {
 		resCb:   nil,
 	}
 	cli.BaseService = *cmn.NewBaseService(nil, "socketClient", cli)
-
+	// FIXME we are loosing "Starting socketClient" message here
+	// add logger to params?
```
melekes: add logger to params or extract Start method?

melekes: https://github.com/tendermint/tendermint/issues/486

melekes: https://github.com/tendermint/abci/issues/80

PR #79, opened 2017-05-01
Update grpc_server.go

faddat:
Added some rudimentary docs.

PR #82, opened 2017-05-09
Update grpc_server.go - 2

adrianbrink:
I've implemented the changes requested in the origin #79


melekes:
I may be a good idea to have comments match the style of ones in golang stdlib (e.g. https://golang.org/pkg/net/rpc/jsonrpc/)

NewSmth returns a new Smth to ...

PR #83, opened 2017-05-09
Test suite for evaluating implementations for bounty

adrianbrink:


adrianbrink:
@ebuchman I will improve the test cases in the future, but I think that this will already be useful for developers wanting to write their own ABCI server implementation.
I've added a small paragraph at the bottom explaining it.

cmd/abci-cli/abci-cli.go
```diff
@@ -177,6 +183,39 @@ func persistentArgs(line []byte) []string {
 
 //--------------------------------------------------------------------------------
 
+func cmdTest(app *cli.App, c *cli.Context) error {
```
ebuchman: How do we know programmatically if the test failed ?

README.md
```diff
@@ -185,3 +185,7 @@ Multiple example apps are included:
 - the `counter` application, which illustrates nonce checking in txs
 - the `dummy` application, which illustrates a simple key-value merkle tree
 - the `dummy --persistent` application, which augments the dummy with persistence and validator set changes
+
+When you are developing an implementation of the ABCI server in your favourite language, we provide the `abci-cli test`
+command that allows you to run our tests against your own ABCI server. Please make sure that your server supports both
+gRPC and the TSP socket protocol.
```
KrzysiekJ: Why is it required? One method is sufficient for an ABCI server to work. Also gRPC is slower than raw socket protocol and may not be available in every language. If there’s a TSP implementation, it’s hard to see an use case for running server using gRPC.

ebuchman: Agreed! We only really need TSP. gRPC is just a convenience and not a requirement, especially since its not even supported yet in every language :)

zramsay:
replaced by #118

PR #84, opened 2017-05-13
Extend the query command

adrianbrink:
Exposes all fields for QueryTx

### DIFF
--- a/cmd/abci-cli/abci-cli.go
+++ b/cmd/abci-cli/abci-cli.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
 
 	abcicli "github.com/tendermint/abci/client"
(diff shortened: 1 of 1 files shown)


cmd/abci-cli/abci-cli.go
```diff
@@ -307,18 +308,36 @@ func cmdCommit(c *cli.Context) error {
 // TODO: Make request and response support all fields.
 func cmdQuery(c *cli.Context) error {
 	args := c.Args()
-	if len(args) != 1 {
-		return errors.New("Command query takes 1 argument")
+	fmt.Println(len(args))
```
melekes: debug?

cmd/abci-cli/abci-cli.go
```diff
@@ -307,18 +308,36 @@ func cmdCommit(c *cli.Context) error {
...
 	}
+
+	var path = "/store"
+	if len(args) > 1 {
+		path = args[1]
+	}
+
+	var height uint64
+	if len(args) > 2 {
+		height, _ = strconv.ParseUint(args[2], 10, 64)
```
melekes: why not check and return an err?

adrianbrink: It isn't an error condition of some of the other parameters aren't given. If they aren't given I just want to fill in the default values.

melekes: No, I mean when user mis inputs something. like `10f` instead of `10`. i.e. checking the result of ParseUInt

cmd/abci-cli/abci-cli.go
```diff
@@ -307,18 +308,36 @@ func cmdCommit(c *cli.Context) error {
 // TODO: Make request and response support all fields.
 func cmdQuery(c *cli.Context) error {
 	args := c.Args()
-	if len(args) != 1 {
-		return errors.New("Command query takes 1 argument")
+
+	if len(args) == 0 {
+		return errors.New("Command query takes 1 or more arguments")
```
ebuchman: This message should say what the args are and their order

PR #87, opened 2017-05-13
Abci result error type comparison

### LINKED ISSUE #85: Simpler ABCI result error type comparison
It would be awesome to be able to compare between error types with code like:
```
res.Is(abci.ErrBaseUnknownAddress)
```
instead of 
```
res.Code == abci.CodeType_BaseUnknownAddress 
```
normally applications don't really ever touch the CodeType unless its doing something fundamental and specialized such as IBC, OR they need to compare error types like above. I think it would just be a bit more straightforward to need to interface with the CodeType at all... thoughts?

rigelrozanski:
closes #85

### DIFF
--- a/types/result.go
+++ b/types/result.go
@@ -29,6 +29,10 @@ func (res Result) IsErr() bool {
 	return res.Code != CodeType_OK
 }
 
+func (res Result) IsSameCode(compare Result) bool {
+	return res.Code == compare.Code
+}
+
 func (res Result) Error() string {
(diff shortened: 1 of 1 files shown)


types/result.go
```diff
@@ -29,6 +29,10 @@ func (res Result) IsErr() bool {
 	return res.Code != CodeType_OK
 }
 
+func (res Result) IsLike(compare Result) bool {
```
ebuchman: Maybe some better name like `IsEqualCode`  or `SameCode` or `IsCode` ? ::shaves yak::

PR #105, opened 2017-09-05
glide: pin tendermint libraries to versions

### LINKED ISSUE #67: Update glide.yaml to point to versioned releases for the tendermint libraries
This can cause dependency conflicts in other packages, think it makes sense to try to update the dependent master branches from their respective develop branches and upgrade acbi here.

zramsay:
- uses latest published version for each library
- closes #67

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -5,15 +5,15 @@ import:
   - proto
 - package: github.com/pkg/errors
 - package: github.com/tendermint/go-crypto
-  version: develop
+  version: v0.2.0
 - package: github.com/tendermint/go-wire
-  version: develop
+  version: v0.6.2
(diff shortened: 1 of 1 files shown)


ebuchman:
I think we only want to do this when we go to master

adrianbrink:
For clarification.

I think this should happen during release. When we want to create a new release, we create a `release-xxx` branch of develop. Then we bump all the version numbers, make any last minute changes and pin the dependencies to a specific version. Then we merge the release branch into both master and develop. On master we tag the commit for future reference.

The only thing is that it might be slightly redundant, since it only prevents someone who is checking out the commit from not being able to fuck up with `glide up`. Generally the `glide.lock` file should be used to ensure consistency.

PR #108, opened 2017-09-21
use request structs for InitChain and BeginBlock

ebuchman:


types/messages.go
```diff
@@ -55,15 +55,15 @@ func ToRequestQuery(reqQuery RequestQuery) *Request {
...
-func ToRequestInitChain(validators []*Validator) *Request {
+func ToRequestInitChain(reqInitChain RequestInitChain) *Request {
 	return &Request{
-		Value: &Request_InitChain{&RequestInitChain{validators}},
+		Value: &Request_InitChain{&reqInitChain},
 	}
 }
 
-func ToRequestBeginBlock(hash []byte, header *Header) *Request {
+func ToRequestBeginBlock(reqBeginBlock RequestBeginBlock) *Request {
```
melekes: I think it can be just `req`

PR #113, opened 2017-10-12
WIP: demo historical queries

melekes:
See https://github.com/tendermint/tendermint/issues/482

### DIFF
--- a/example/dummy/dummy.go
+++ b/example/dummy/dummy.go
@@ -4,15 +4,14 @@ import (
 	"strings"
 
 	"github.com/tendermint/abci/types"
-	"github.com/tendermint/merkleeyes/iavl"
+	"github.com/tendermint/iavl"
 	cmn "github.com/tendermint/tmlibs/common"
-	"github.com/tendermint/tmlibs/merkle"
 )
 
 type DummyApplication struct {
 	types.BaseApplication
 
-	state merkle.Tree
+	state *iavl.IAVLTree
 }
 
 func NewDummyApplication() *DummyApplication {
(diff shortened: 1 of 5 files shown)


example/dummy/dummy.go
```diff
@@ -46,12 +45,12 @@ func (app *DummyApplication) Commit() types.Result {
...
 func (app *DummyApplication) Query(reqQuery types.RequestQuery) (resQuery types.ResponseQuery) {
 	if reqQuery.Prove {
-		value, proof, exists := app.state.Proof(reqQuery.Data)
+		value, proof, err := app.state.GetWithProof(reqQuery.Data)
 		resQuery.Index = -1 // TODO make Proof return index
 		resQuery.Key = reqQuery.Data
 		resQuery.Value = value
-		resQuery.Proof = proof
-		if exists {
+		resQuery.Proof = proof.Bytes()
```
odeke-em: In the else condition i.e when `err != nil` I wonder if instead the iavl library could pass back a sentinel error that we can then explicitly check e.g
`if err == iavl.ErrNotExists` because perhaps GetWithProof encounters another error for which this query should entirely abort.

Just a thought though.

example/dummy/persistent_dummy.go
```diff
@@ -87,7 +87,8 @@ func (app *PersistentDummyApplication) CheckTx(tx []byte) types.Result {
 
 func (app *PersistentDummyApplication) Commit() types.Result {
 	// Save
-	appHash := app.app.state.Save()
+	// appHash := app.app.state.Save()
```
odeke-em: Minor nit: let's remove this commented code out.

melekes:
Waiting until the IAVL interface stabilizes https://github.com/tendermint/iavl/pull/8

ebuchman:
Thanks, we've taken this up again in https://github.com/tendermint/abci/pull/116/commits

Issue #114, opened 2017-10-16
client/socketClient: deadlock when we invoke StopForError due to incorrect mutex use


odeke-em:
I was just planning on writing a malicious ABCI client for security hardening, and I noticed a nice simple way in which the server/mainnet can potentially choose to cheat clients and maliciously paint them as "Byzantine", because of a deadlock in the StopOnError, by sending them malformed bytes that don't parse

https://github.com/tendermint/abci/blob/15cd7fb1e3b75c436b6dee89a44db35f3d265bd0/client/socket_client.go#L100-L104

Notice that we lock it but never release the mutex, lock that two times and kaboom, that's a deadlock and the client won't respond to any other actions. Please see a simple repro here
https://play.golang.org/p/jsgb50T62o
or inlined
```go
package main

import (
	"errors"
	"sync"
)

type client struct {
	mu  sync.Mutex
	err error
}

func (c *client) StopOnError(err error) {
	c.mu.Lock()
	if !c.IsRunning() {
		return
	}

	if err != nil {
		c.err = err
	}
	c.mu.Unlock()
}

func (c *client) IsRunning() bool {
	return false
}

func main() {
	c := new(client)
	err := errors.New("foo")
	c.StopOnError(err)
	c.StopOnError(err)
}
```

I belive we meant to instead make it
```diff
--- orig.go	2017-10-16 17:25:48.000000000 -0600
+++ patched.go	2017-10-16 17:26:31.000000000 -0600
@@ -11,15 +11,15 @@
 }
 
 func (c *client) StopOnError(err error) {
-	c.mu.Lock()
 	if !c.IsRunning() {
 		return
 	}
 
 	if err != nil {
+		c.mu.Lock()
 		c.err = err
+		c.mu.Unlock()
 	}
-	c.mu.Unlock()
 }
 
 func (c *client) IsRunning() bool {
```

melekes:
@odeke-em nice catch
![](https://media.giphy.com/media/WUexQ89aMP0re/giphy.gif)

odeke-em:
haha, thanks @melekes for the GIF.

odeke-em:
Alright, the PR for it is https://github.com/tendermint/abci/pull/115

PR #115, opened 2017-10-18
socketClient: fix and test for StopForError deadlock

odeke-em:
Fixes https://github.com/tendermint/abci/issues/114.

Fix and test for StopForError deadlock due to accidental
Mutex misuse. The test should return instantenously if
StopForError works without erraneous contention, lest it regressed.


ebuchman:
Wow, nice find

odeke-em:
Thanks for the reviews and merge y'all.

PR #116, opened 2017-10-18
merkleeyes to iavl

ethanfrey:
Rip out all traces of merkleeyes.
Use iavl.VersionedTree with optional persistence for all operations.

This lets us use the dummy app with new code for test suites, among other things.

### DIFF
--- a/example/dummy/dummy.go
+++ b/example/dummy/dummy.go
@@ -4,19 +4,20 @@ import (
 	"strings"
 
 	"github.com/tendermint/abci/types"
-	"github.com/tendermint/merkleeyes/iavl"
+	wire "github.com/tendermint/go-wire"
+	"github.com/tendermint/iavl"
 	cmn "github.com/tendermint/tmlibs/common"
-	"github.com/tendermint/tmlibs/merkle"
+	dbm "github.com/tendermint/tmlibs/db"
 )
 
 type DummyApplication struct {
 	types.BaseApplication
 
-	state merkle.Tree
+	state *iavl.VersionedTree
 }
 
 func NewDummyApplication() *DummyApplication {
-	state := iavl.NewIAVLTree(0, nil)
+	state := iavl.NewVersionedTree(0, dbm.NewMemDB())
 	return &DummyApplication{state: state}
 }
 
@@ -46,22 +47,26 @@ func (app *DummyApplication) Commit() types.Result {
 
 func (app *DummyApplication) Query(reqQuery types.RequestQuery) (resQuery types.ResponseQuery) {
 	if reqQuery.Prove {
-		value, proof, exists := app.state.Proof(reqQuery.Data)
+		value, proof, err := app.state.GetWithProof(reqQuery.Data)
+		// be stupid here
+		if err != nil {
+			panic(err)
+		}
 		resQuery.Index = -1 // TODO make Proof return index
 		resQuery.Key = reqQuery.Data
 		resQuery.Value = value
-		resQuery.Proof = proof
-		if exists {
+		resQuery.Proof = wire.BinaryBytes(proof)
+		if value == nil {
 			resQuery.Log = "exists"
 		} else {
 			resQuery.Log = "does not exist"
 		}
 		return
 	} else {
-		index, value, exists := app.state.Get(reqQuery.Data)
+		index, value := app.state.Get(reqQuery.Data)
 		resQuery.Index = int64(index)
 		resQuery.Value = value
-		if exists {
+		if value != nil {
 			resQuery.Log = "exists"
 		} else {
 			resQuery.Log = "does not exist"
--- a/example/dummy/dummy.go
+++ b/example/dummy/dummy.go
@@ -1,7 +1,6 @@
 package dummy
 
 import (
-	"fmt"
 	"strings"
 
 	"github.com/tendermint/abci/types"
@@ -34,7 +33,6 @@ func (app *DummyApplication) DeliverTx(tx []byte) types.Result {
 	} else {
 		app.state.Set(tx, tx)
 	}
-	fmt.Println("set data")
 	return types.OK
 }
 
@@ -63,7 +61,7 @@ func (app *DummyApplication) Commit() types.Result {
 func (app *DummyApplication) Query(reqQuery types.RequestQuery) (resQuery types.ResponseQuery) {
 	if reqQuery.Prove {
 		value, proof, err := app.state.GetWithProof(reqQuery.Data)
-		// be stupid here
+		// if this wasn't a dummy app, we'd do something smarter
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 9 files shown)


tests/test_cli/test.sh
```diff
@@ -15,6 +15,7 @@ function testExample() {
 
 	echo "Example $N"
 	$APP &> /dev/null &
+    # $APP &> ./app.out &
```
melekes: m?

example/dummy/dummy.go
```diff
@@ -32,6 +34,7 @@ func (app *DummyApplication) DeliverTx(tx []byte) types.Result {
 	} else {
 		app.state.Set(tx, tx)
 	}
+	fmt.Println("set data")
```
melekes: did you use it for debug or we need this?

example/dummy/dummy.go
```diff
@@ -40,28 +43,45 @@ func (app *DummyApplication) CheckTx(tx []byte) types.Result {
 }
 
 func (app *DummyApplication) Commit() types.Result {
-	hash := app.state.Hash()
+	// Save a new version
+	var hash []byte
+	var err error
+
+	if app.state.Size() > 0 {
```
melekes: do we need this condition? I mean, by the time we commit the size should always be > 0, no?

example/dummy/dummy.go
```diff
@@ -40,28 +43,45 @@ func (app *DummyApplication) CheckTx(tx []byte) types.Result {
...
+	}
+
 	return types.NewResultOK(hash, "")
 }
 
 func (app *DummyApplication) Query(reqQuery types.RequestQuery) (resQuery types.ResponseQuery) {
 	if reqQuery.Prove {
-		value, proof, exists := app.state.Proof(reqQuery.Data)
+		value, proof, err := app.state.GetWithProof(reqQuery.Data)
+		// be stupid here
```
melekes: I'd prefer `if this wasn't a dummy app, we'd do something smarter` here too

example/dummy/dummy.go
```diff
@@ -40,28 +43,45 @@ func (app *DummyApplication) CheckTx(tx []byte) types.Result {
...
+		if err != nil {
+			panic(err)
+		}
 		resQuery.Index = -1 // TODO make Proof return index
 		resQuery.Key = reqQuery.Data
 		resQuery.Value = value
-		resQuery.Proof = proof
-		if exists {
+		resQuery.Proof = wire.BinaryBytes(proof)
+		if value == nil {
```
melekes: shouldn't this be `if value != nil { "exists"`

example/dummy/persistent_dummy.go
```diff
@@ -86,18 +89,22 @@ func (app *PersistentDummyApplication) CheckTx(tx []byte) types.Result {
...
-		Height:  app.blockHeader.Height,
-		AppHash: appHash, // this hash will be in the next block header
-	}
+	// Save a new version
+	var appHash []byte
+	var err error
 
-	app.logger.Info("Saving block", "height", lastBlock.Height, "root", lastBlock.AppHash)
-	SaveLastBlock(app.db, lastBlock)
+	if app.app.state.Size() > 0 {
```
melekes: same comment as above

example/dummy/persistent_dummy.go
```diff
@@ -86,18 +89,22 @@ func (app *PersistentDummyApplication) CheckTx(tx []byte) types.Result {
 }
 
 func (app *PersistentDummyApplication) Commit() types.Result {
-	// Save
-	appHash := app.app.state.Save()
-	app.logger.Info("Saved state", "root", appHash)
+	app.height = app.blockHeader.Height
```
melekes: shouldn't we do this after we successfully saved a new version?

example/dummy/dummy.go
ethanfrey: Debug. Oops

example/dummy/dummy.go
ethanfrey: Well. We have tests that commit a few empty blocks before adding a TX. They were returning errors here, which broke things. So I added tests. Thus the slack ping to alexis

example/dummy/dummy.go
ethanfrey: Yes. Good eye

example/dummy/persistent_dummy.go
ethanfrey: I guess so, but if save fails, we panic, and this is only stored locally, so it doesn't really matter.
But sure, I'll clean it up.

example/dummy/dummy.go
```diff
@@ -40,28 +41,45 @@ func (app *DummyApplication) CheckTx(tx []byte) types.Result {
...
 	return types.NewResultOK(hash, "")
 }
 
 func (app *DummyApplication) Query(reqQuery types.RequestQuery) (resQuery types.ResponseQuery) {
 	if reqQuery.Prove {
-		value, proof, exists := app.state.Proof(reqQuery.Data)
+		value, proof, err := app.state.GetWithProof(reqQuery.Data)
+		// if this wasn't a dummy app, we'd do something smarter
+		if err != nil {
+			panic(err)
```
ebuchman: does this panic if not found

example/dummy/dummy.go
ebuchman: Hm. For the persistent dummy I ended up adding the constraint that InitChain must be called, which will make the state non-zero. In general would be nice if we could still get versions of empty trees so we dont have to worry about this edge condition

cloudhead: We can use `!app.state.IsEmpty()` now which is a bit cleaner, but if a tree has no root, what are we saving exactly? We could potentially have it just save a version number but hmmm..

example/dummy/persistent_dummy.go
```diff
@@ -36,17 +35,22 @@ type PersistentDummyApplication struct {
...
-	stateTree := iavl.NewIAVLTree(0, db)
-	stateTree.Load(lastBlock.AppHash)
+	db, err := dbm.NewGoLevelDB(name, dbDir)
+	if err != nil {
+		panic(err)
+	}
 
-	// log.Notice("Loaded state", "block", lastBlock.Height, "root", stateTree.Hash())
+	stateTree := iavl.NewVersionedTree(500, db)
+	if !empty {
```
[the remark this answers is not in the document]
ethanfrey: yes, this is possible now, with fixes by alexis. I will update to the newest iavl develop and then remove this check.

example/dummy/persistent_dummy.go
```diff
@@ -57,9 +61,8 @@ func (app *PersistentDummyApplication) SetLogger(l log.Logger) {
 
 func (app *PersistentDummyApplication) Info(req types.RequestInfo) (resInfo types.ResponseInfo) {
 	resInfo = app.app.Info(req)
-	lastBlock := LoadLastBlock(app.db)
-	resInfo.LastBlockHeight = lastBlock.Height
-	resInfo.LastBlockAppHash = lastBlock.AppHash
+	resInfo.LastBlockHeight = app.height
+	resInfo.LastBlockAppHash = app.app.state.Hash()
```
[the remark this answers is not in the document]
ethanfrey: app.state.Hash() is the root hash of the last committed tree, given the new versionedtree functionality. which was the same as lastBlock, no?

example/dummy/dummy.go
ebuchman: It's just for useability. The tree is still something the app thinks about and has a root (even if empty). And for recover logic, we like the tree version to equal the height, even if the tree hasn't changed.

PR #117, opened 2017-10-18
use spf13/cobra instead of urfave/cli

zramsay:
- closes #95, #101

PR #118, opened 2017-10-19
Add test command to abci-cli

### LINKED ISSUE #76: Test suite for evaluating implementations for bounty
There's some scripts that test eg. the "first abci guide" stuff. We should abstract this so it's easier to run on new implementations. Also, we should add an integration test to see it run with tendermint.

A successful implementation should come with a docker file so the server and app can be easily started. We can then run a series of txs and ensure things act as expected.

@adrianbrink want to work on this?

KrzysiekJ:
This supersedes #83 (commits squashed with slight changes + a simple fix) and fixes #76.


tests/server/client.go
```diff
@@ -0,0 +1,103 @@
...
+	res := client.CommitSync()
+	_, data, log := res.Code, res.Data, res.Log
+	if res.IsErr() {
+		fmt.Println("Failed test: Commit")
+		fmt.Printf("committing %v\nlog: %v", log)
+		return errors.New(res.Error())
+	}
+	if !bytes.Equal(data, hashExp) {
+		fmt.Println("Failed test: Commit")
+		fmt.Printf("Commit hash was unexpected. Got %X expected %X",
```
KrzysiekJ: If this test fails, the first value is incorrectly rendered. For example, if the second value is `0000000000000005` and the first one is the same, but with 0 byte added at the end, we get “Got 303030303030303030303030303030353030 expected 0000000000000005”. Maybe it has something to do with the fact that the first value is `data.Bytes`, while the second is `[]byte`?

tests/server/client.go
```diff
@@ -0,0 +1,103 @@
...
+			data, hashExp)
+		return errors.New("CommitTx failed")
+	}
+	fmt.Println("Passed test: Commit")
+	return nil
+}
+
+func DeliverTx(client abcicli.Client, txBytes []byte, codeExp types.CodeType, dataExp []byte) error {
+	res := client.DeliverTxSync(txBytes)
```
melekes: what about checking the `res.IsErr()`? similar to `CheckTx`

cmd/abci-cli/abci-cli.go
```diff
@@ -196,6 +202,45 @@ func persistentArgs(line []byte) []string {
...
+	err = servertest.DeliverTx(client, []byte{0x00, 0x03}, types.CodeType_OK, nil)
+	fmt.Println("")
+	err = servertest.DeliverTx(client, []byte{0x00, 0x00, 0x04}, types.CodeType_OK, nil)
+	fmt.Println("")
+	err = servertest.DeliverTx(client, []byte{0x00, 0x00, 0x06}, types.CodeType_BadNonce, nil)
+	fmt.Println("")
+	err = servertest.Commit(client, []byte{0, 0, 0, 0, 0, 0, 0, 5})
+
+	if err != nil {
+		return errors.New("Some checks didn't pass, please use the cli to see the exact failures.")
```
melekes: huh? this command is a cli command. what do you mean by `use the cli`?

tests/server/client.go
melekes: have you tried `data.Bytes()`?

cmd/abci-cli/abci-cli.go
```diff
@@ -143,14 +143,20 @@ func main() {
 				},
 			},
 		},
+		{
+			Name:  "test",
+			Usage: "Runs the integration tests",
+			Action: func(c *cli.Context) error {
+				return cmdTest(app, c)
+			},
+		},
```
zramsay: we're switching to cobra in #117, so this'll need to be re-worked. Maybe base this PR on that one?

cmd/abci-cli/abci-cli.go
zramsay: i think he means "debug manually", like if in running `abci-cli test` there's a failure at one of the lines, use abci-cli deliver_tx` or whatever to debug. Instead, printing the actual `err` message on this line would show the exact failures, would it not?

tests/server/client.go
KrzysiekJ: Looks like `res.IsErr()` checks whether code is not `OK`. However we already have the `codeExp` argument. I guess that the check of `res.IsErr()` should be removed from `CheckTx` then, as it is redundant and will lead to mismatch if the caller expects something different than `OK`.

tests/server/client.go
KrzysiekJ: `data.Bytes()` helps, thanks.

cmd/abci-cli/abci-cli.go
KrzysiekJ: I’ll rebase if #117 is merged earlier.

cmd/abci-cli/abci-cli.go
KrzysiekJ: I’ll change “use the cli” to “inspect stdout”.

cmd/abci-cli/abci-cli.go
```diff
@@ -196,6 +202,45 @@ func persistentArgs(line []byte) []string {
...
+	fmt.Println("")
+	err = servertest.DeliverTx(client, []byte{0x00, 0x03}, types.CodeType_OK, nil)
+	fmt.Println("")
+	err = servertest.DeliverTx(client, []byte{0x00, 0x00, 0x04}, types.CodeType_OK, nil)
+	fmt.Println("")
+	err = servertest.DeliverTx(client, []byte{0x00, 0x00, 0x06}, types.CodeType_BadNonce, nil)
+	fmt.Println("")
+	err = servertest.Commit(client, []byte{0, 0, 0, 0, 0, 0, 0, 5})
+
+	if err != nil {
```
KrzysiekJ: @zramsay: Haven’t noticed your comment (visible now in the outdated diff). It looks like there may be many errors, assigned in many checks and `err` in this line is only the last one (or `nil`), so I’m not sure whether previous assignments have any effect. Actually, if there are some errors, but the last commit test is correct, then it looks like this function will return `nil`… We should therefore probably change each `err = […]` assignment to something like `err = err || […]`.

KrzysiekJ: I’ve updated the commit, introducing a helper `or` function.

melekes:
@KrzysiekJ sorry it took so long to respond. could you please rebase it against develop. let's merge this!

KrzysiekJ:
Rebased. :)

README.md
```diff
@@ -187,3 +187,47 @@ Here, we describe the requests and responses as function arguments and return va
...
+## In Process
+
+The simplest implementation just uses function calls within Go.
+This means ABCI applications written in Golang can be compiled with TendermintCore and run as a single binary.
+
+
+# Tools and Apps
+
+The `abci-cli` tool wraps any ABCI client and can be used for probing/testing an ABCI application.
+See the [guide](https://tendermint.com/docs/guides/abci-cli) for more details.
```
melekes: https://tendermint.readthedocs.io/en/master/abci-cli.html

Makefile
```diff
@@ -8,8 +8,8 @@ all: protoc install test
 NOVENDOR = go list github.com/tendermint/abci/... | grep -v /vendor/
 
 install-protoc:
-	# Download: https://github.com/google/protobuf/releases
-	go get github.com/golang/protobuf/protoc-gen-go
+# Download: https://github.com/google/protobuf/releases
```
melekes: ?

KrzysiekJ: This was in the original PR. I’ve reverted that line (will push in a moment).

KrzysiekJ:
I’ve resolved conflict with `develop` and updated the code to compile with recent changes. Hope it can be merged now.

melekes:
`tests/test_cli/test.sh 100644 → 100755`

melekes:
Thank you very much for working on this! Hopefully, @ebuchman will approve and merge it

cmd/abci-cli/abci-cli.go
```diff
@@ -271,6 +274,16 @@ var dummyCmd = &cobra.Command{
 	},
 }
 
+var testCmd = &cobra.Command{
+	Use:   "test",
+	Short: "Run integration tests",
+	Long:  "",
```
zramsay: can we get a long description here? Even if it's the same as Short, they show up differently

cmd/abci-cli/abci-cli.go
```diff
@@ -287,6 +300,53 @@ func persistentArgs(line []byte) []string {
...
+	err = or(err, servertest.DeliverTx(client, []byte{0x01}, code.CodeTypeOK, nil))
+	fmt.Println("")
+	err = or(err, servertest.DeliverTx(client, []byte{0x00, 0x02}, code.CodeTypeOK, nil))
+	fmt.Println("")
+	err = or(err, servertest.DeliverTx(client, []byte{0x00, 0x03}, code.CodeTypeOK, nil))
+	fmt.Println("")
+	err = or(err, servertest.DeliverTx(client, []byte{0x00, 0x00, 0x04}, code.CodeTypeOK, nil))
+	fmt.Println("")
+	err = or(err, servertest.DeliverTx(client, []byte{0x00, 0x00, 0x06}, code.CodeTypeBadNonce, nil))
+	fmt.Println("")
```
zramsay: can these `fmt.Println("")` be handled in `func or()` ?

melekes: I'd prefer not to.

cmd/abci-cli/abci-cli.go
ebuchman: None of the others have it. Feel free to add them tho :)

PR #120, opened 2017-10-25
Check for a two part, protocol provided, addr

### LINKED ISSUE #72: Validate client address arg
client, err := abciclient.NewClient("127.0.0.1:46657", "socket", true)

```
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/tendermint/tendermint/vendor/github.com/tendermint/go-common.Connect(0x87bdd3, 0xf, 0x4470ed, 0xc4200b7be8, 0x4470ed, 0x0)
        /home/vagrant/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/go-common/net.go:11 +0xc5
github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/client.(*socketClient).OnStart(0xc4200aefa0, 0x60, 0xc42005e4e0)
        /home/vagrant/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/client/socket_client.go:68 +0x3e
github.com/tendermint/tendermint/vendor/github.com/tendermint/go-common.(*BaseService).Start(0xc4200aefa0, 0x0, 0xc42005e4e0, 0xc420016400)
        /home/vagrant/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/go-common/service.go:96 +0x2c2
github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/client.NewSocketClient(0x87bdd3, 0xf, 0x4af201, 0xa6b4a0, 0xc42000e018, 0xc4200b7ea8)
        /home/vagrant/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/client/socket_client.go:57 +0x23c
github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/client.NewClient(0x87bdd3, 0xf, 0x876c8d, 0x6, 0x1, 0x0, 0xc420011630, 0xc420011600, 0xc420011600)
        /home/vagrant/go/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/abci/client/client.go:49 +0x17a
main.main()
        /home/vagrant/go/src/github.com/tendermint/tendermint/transacter.go:21 +0x1bf
exit status 2
```

`client, err := abciclient.NewClient("tcp://127.0.0.1:46657", "socket", true)` works.

Return an error?

ericdmann:
Closes #72 

Since the included lib (go-common) is deprecated, and updated tmlibs contains the proper check, handle here until new tmlibs can be used exclusively.

https://github.com/tendermint/go-common/blob/47e06734f6ee488cc2e61550a38642025e1d4227/net.go#L11

### DIFF
--- a/client/socket_client.go
+++ b/client/socket_client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
@@ -59,6 +60,10 @@ func NewSocketClient(addr string, mustConnect bool) *socketClient {
 func (cli *socketClient) OnStart() error {
 	if err := cli.BaseService.OnStart(); err != nil {
 		return err
+	}
+
+	if len(strings.SplitN(cli.addr, "://", 2)) < 2 {
+		return fmt.Error("Missing protocol provided (ex. tcp://)", cli.addr)
 	}
 
 	var err error
--- a/client/socket_client.go
+++ b/client/socket_client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
@@ -57,6 +58,10 @@ func NewSocketClient(addr string, mustConnect bool) *socketClient {
 }
 
 func (cli *socketClient) OnStart() error {
+
+	if len(strings.SplitN(cli.addr, "://", 2)) < 2 {
+		return fmt.Errorf("Missing protocol (ex. tcp://) in %v", cli.addr)
+	}
 
 	if err := cli.BaseService.OnStart(); err != nil {
 		return err
--- a/client/socket_client.go
+++ b/client/socket_client.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"reflect"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


client/socket_client.go
```diff
@@ -61,6 +62,10 @@ func (cli *socketClient) OnStart() error {
 		return err
 	}
 
+	if len(strings.SplitN(cli.addr, "://", 2)) < 2 {
+		return fmt.Error("Missing protocol provided (ex. tcp://)", cli.addr)
```
melekes: Maybe `Missing protocol (ex. tcp://) in %v`?

melekes:
@ericdmann thank you! Could you please also write a small test for this?

client/socket_client.go
ericdmann: Love it, will incorporate

ericdmann:
Test added - excuse the extra commit, vendor copy stole the change for a moment :)

ebuchman:
Thanks for this, but not sure I understand. `addr` is only used in `cmn.Connect`, which defaults to `tcp://` if not provided. Why do we need this?

ebuchman:
That said, you've inspired me to do https://github.com/tendermint/abci/commit/6c41ec65bcd86489041fb3bf3a7c56cbe9e70dcf

ericdmann:
The version of Connect() used here (if I'm branched correctly) is from the deprecated go-common (https://github.com/tendermint/go-common/blob/47e06734f6ee488cc2e61550a38642025e1d4227/net.go#L11) and does not check size or default to tcp://, which seems to be at the source of #72. 

This change will protect against the deprecated version until the new tmlibs `Connect` can be used here.

melekes:
Shouldn't we rewrite abci to use tmlibs version then?

ebuchman:
> Shouldn't we rewrite abci to use tmlibs version then?

As far as I can tell it does?

ericdmann:
@ebuchman you're correct, I was working off the trace in the #72 description -- but yes, `tmlibs` solves.

PR #121, opened 2017-10-25
Add proposer to block header

KrzysiekJ:
Refs tendermint/tendermint#338.


types/types.proto
```diff
@@ -207,6 +207,7 @@ message Header {
 	bytes data_hash = 7;
 	bytes validators_hash = 8;
 	bytes app_hash = 9;
+        bytes proposer = 10;
```
KrzysiekJ: It is a bit problematic what to actually put in this field:

* A public key? If so, we lose information about key type.
* An address? It encompasses both key type and the key itself, but they are encoded in a non-reversible hash.

melekes:
https://github.com/tendermint/tendermint/issues/338#issuecomment-297078166

ebuchman:
Thanks - closing for now, will deal with shortly via https://github.com/tendermint/abci/issues/152

PR #123, opened 2017-11-06
Add counter binary to demo application fully

ethanfrey:
Gautier was looking at the abci repo to follow up from Adrian's talk and understand the abci structure.

While there are some example apps, there is no code to show how to actually run them. Nor are they commented.

I added a simple, commented main package for counter, which one can connect to via abci-cli. I thought there were apps like this before. We definitely need better documentation here if we point new developers to this package.

### DIFF
--- a/cmd/counter/main.go
+++ b/cmd/counter/main.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+
+	cmn "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tmlibs/log"
+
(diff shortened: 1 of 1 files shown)


zramsay:
https://github.com/tendermint/abci/blob/master/cmd/abci-cli/abci-cli.go#L418-L440

zramsay:
arguably the read the docs doc should be rendered from this repo's readme.rst, like we do with tools

ebuchman:
Closing for https://github.com/tendermint/abci/issues/134

zramsay:
@melekes yes, doing that now. Will add to https://github.com/tendermint/tendermint/pull/1046

Issue #125, opened 2017-11-08
Connection error module=abci-server error=EOF


paulrd:
I get this error while running the dummy or counter app and sending messages via the cli.

melekes:
Could you please share the logs or sequence of steps?

paulrd:
1. In first teminal launch abci-cli counter
2. In another terminal run abci-cli check_tx 0x00
3. In first terminal output is:
```bash
I[11-09|12:48:26.882] Starting ABCIServer                          module=abci-server impl=ABCIServer
I[11-09|12:48:26.882] Waiting for new connection...                module=abci-server 
I[11-09|12:48:51.347] Accepted a new connection                    module=abci-server 
I[11-09|12:48:51.347] Waiting for new connection...                module=abci-server 
E[11-09|12:48:51.350] Connection error                             module=abci-server error=EOF
```
The counter app and abci-cli otherwise perform as intended, so I'm wondering if this connection error is also intended. Looking at the code, it seems that reading from the socket is not blocking as I would expect while there is not a message yet. It probably reads the first message ok, then looks for the next one and gets EOF unstead of a blocking read. I don't think I understand the intended functionality.

paulrd:
Ok, I think I see the problem. The call to types.ReadMessage [here](https://github.com/tendermint/abci/blob/dc33aad9b4e514a2322725ef68f27f72d955c537/server/socket_server.go#L147) should be a blocking call or at least block somewhere along the call stack from there (ReadByteSlice->ReadVarint->ReadUInt8->ReadByte->ReadFull). Possibly by just calling ReadAt instead of ReadFull at the appropriate time so that we get a blocking read instead of EOF when the client has not yet closed the connection but no data is waiting. This will be performant as the call to ReadMessage is already in a go routine. At cursory glance, this may be a bug that could cause a lot of EOF signals to come on the channel when they should not.

paulrd:
I think I'm wrong. There shouldn't be any 'non-blocking' read functions. I'm having a hard time getting the definitive answer looking around the net.

paulrd:
In any case, even with pull request #129, the abci-cli dummy and counter apps close their connections after every command despite the abci documentation saying otherwise. I'm new to golang and haven't been able to figure out why.

melekes:
the abci-cli dummy and counter apps do not close their connections. rather `abci-cli console` or `abci-cli check_tx`, etc. do after every command. this is not really an error per se. we should probably silent the logger in this case.

[...]

PR #129, opened 2017-11-09
Wrong 'error' variable is being read in socket_server.go

paulrd:
This is likely related to #125

### DIFF
--- a/server/socket_server.go
+++ b/server/socket_server.go
@@ -117,7 +117,7 @@ func (s *SocketServer) acceptConnectionsRoutine() {
 		go func() {
 			// Wait until signal to close connection
 			errClose := <-closeConn
-			if err == io.EOF {
+			if errClose == io.EOF {
 				s.Logger.Error("Connection was closed by client")
 			} else if errClose != nil {
 				s.Logger.Error("Connection error", "error", errClose)


melekes:
not sure about this actually. see 83920a1c373d4e9c91b4f2d0f8ba3a40b8cdf5b0. @ebuchman /cc

paulrd:
Yeah, see my discussion at #125. I think if you had a blocking read (in HandleRequests), you should never get EOF until the connection is closed (by client) so that return statement might make sense. As it is now, you can get EOF when the next message hasn't arrived from the client yet (IIUC).

melekes:
this change breaks the code. if you try running TM and abci-cli, TM will fail with "Connection was closed by client" error. the original code is correct.

paulrd:
That's strange, the only effect this code should have would be in the error message logged. Isn't TM the client? It's the abci-cli (dummy or counter) app (server) that will 'fail' with "Connection was closed by client". EOF gets detected with a different message in the very next clause of the if statement anyway. So instead of "Connection was closed by client" you get "Connection error module=abci-server error=EOF". This pull request doesn't change the logic of the abci-server at all. Checking err instead of errClose is clearly a bug (if not a very severe one). Why would you want to check err here anyway?

ebuchman:
This was clearly a bug. Thanks for finding and fixing it! Sorry for delays/confusion :)

Issue #57, opened 2017-01-19
How to deal with fees/gas in tendermint consensus?


ethanfrey:
A good consensus algorithm needs a way to prioritize multiple transactions, when there is more than a trivial level of transactions.

In order to do this well, the tendermint core will need to be aware of the fees and gas associated with a given transaction.  There are two general ideas how to do this:

1. We no longer broadcast arbitrary bytes, but a special format the core can validate tx, verify sigs, parse the fees and gas, etc.
2. Update check_tx so that abci app validates the tx in it's own way, but then returns the fees/gas of any valid transaction to the core, so it can use that for routing

After some discussion, @jaekwon and me decided that the second approach was more flexible.  We could implement this base logic in basecoin, and people could write plugins to just take advantage of this existing logic. The other approach is a slippery slope, adding more and more logic into tendermint core, until it is no longer a general purpose consensus engine.

The proposal is to make the following changes:

abci interface:
  * `CheckTx` returns two more fields, fees and gas
  * These fields are optional.  If an app doesn't return them, tendermint core will use some defaults
  * Maybe define new error types for validation?

basecoin (or other abci implementation):
  * on `CheckTx`, app validates tx, checks sigs, checks fees and gas, checks the account has the money
    * On success, return gas/fees values to core
    * If any problem (account doesn't have the fees it promises), return an error
  * on `DeliverTx`, app removes 100% of Fees from account regardless of success/fail of app call
    * App specific logic what to do with them (destroy them?  distribute to the validator)
    * This can be configured by passing a `FeeHandler` interface into `NewBasecoin` or a similar approach
    * plugin decides what to do with gas... 

tendermint core:

  * on receive new tx, tendermint calls `CheckTx` (as now)
  * if tx is valid, core remembers the fees/gas associated with it for the consensus round
  * when voting on a block, reject any block with a total gas over the chain-configured gas limit
  * when proposing a block, do some algorithm to maximize the fees while staying within the gas limit
    * (this is a [well known NP-problem](https://en.wikipedia.org/wiki/Knapsack_problem) and we can use some of the approximation algorithms from cs literature) 
    * to limit breakdown of consensus, other nodes will not reject a proposal because they found a better solution

Question from Jae, how do you configure the gas limit?  If it is different between nodes, this will quickly lead to a breakdown in consensus.  Changing it is analogous to changing the maximum blocksize in bitcoin. In other words, quite easy 😉 

As a first step, the gas limit per block can be set inside the genesis block to ensure consistency.

[...]

PR #130, opened 2017-11-09
Tags in DeliverTx response

melekes:


melekes:
> If there are things we know we need numbers for, like gas, we should probably just build those in direct.

my thinking was by declaring a type we're avoiding parsing errors. For example, tag "count_participants": 5. And the user wants to query "count_participants > 3". If value were a string ("5"), we'd have to parse it as Int each time and then compare against the query. If the value is an int, we're avoiding parsing (protobuf ensures it is an int64) and can also save it as int in the DB. But maybe this is not worth adding a new field.

ebuchman:
Interesting. I'm not sure if it's worth it. What does @ethanfrey think?

melekes:
Originally this design was suggested by Jae.

melekes:
```
message OneMessage {
  enum Type { STRING = 1; INT = 2; }
  required Type type = 1;

  optional string ValueString = 2;
  optional int64 ValueInt = 3;
}
```

we can also add a type so we could switch in code ^^^

ethanfrey:
I originally proposed a list of strings and only doing exact matches. thus it would match a specific key-value pair.

This concept was turned much more complex to the string-{string,int} model proposed in a meeting with anton, jae, and bucky.

AFAIK, the only use cases of the queries needed by the API are for exact matches, and if we are not using integer comparison (does anyone really need to filter all tx with a value over 40 atom regardless of address?), I am fine using only string equality (i can encode addresses as strings).

If we do support both int and strings and comparison operators in the indexer, however, it does make sense to expose that in the abci api.

types/result.go
```diff
@@ -102,6 +103,7 @@ func (r *ResponseCheckTx) Result() Result {
 		Code: r.Code,
 		Data: r.Data,
 		Log:  r.Log,
+		Tags: r.Tags,
```
ethanfrey: There is no meaning to tags in checktx, as they should never be indexed. We should remove that from the api, as it would be misleading.

types/result.go
```diff
@@ -12,6 +12,7 @@ type Result struct {
 	Code CodeType   `json:"code"`
 	Data data.Bytes `json:"data"`
 	Log  string     `json:"log"` // Can be non-deterministic
+	Tags []*KVPair  `json:"tags"`
```
ethanfrey: I think we should start separating the result from DeliverTx and CheckTx.
DeliverTx needs tags, which CheckTx doesn't.
We will also need to add Gas and Fees (or Cost and Value or....) to CheckTx as per #57 and soon.

Given this, I already separated the DeliverTx and CheckTx results in the sdk. We should use different types in the abci.

ethanfrey:
I agree with this PR given the design. I still have issues with the design, and the lack of ability to have any meaningful input once after I couldn't make one meeting late in my evening.

The actual needs of the UI team involve matching senders and receivers. As the basecoin design allows for multiple senders, and we only allow one value for a key, we can only match tx where the user was the first sender or first receiver. Or maybe choose some value N where we search for sender1=XYZ sender2=XYZ sender3=XYZ etc...

Actually, the string channel name idea would work easier in this case. DeliverTx could emit "coin.sender=XYZ", "coin.sender=ABC", "coin.sender=DEF" as three different tags and we could subscribe to our tx just the same regardless of how many tx inputs there were. 

But this is just me being a grumpy old man..... given the design space that was agreed upon, I approve this pr, but please use different result types here:
https://github.com/tendermint/abci/blob/feature/tags-in-deliver-tx/types/application.go#L16
and here:
https://github.com/tendermint/abci/blob/feature/tags-in-deliver-tx/types/application.go#L21

melekes:
I don't think it's too late to switch to Redis like subscriptions in that is what we want and it is enough for everyone.

ethanfrey:
Added comments on slack, but they belong here:

We actually don't need any general purpose > < and strings are fine for subscriptions.
We only need to add a special case `height int` that supports `<` and `>` just for search, not subscriptions.

If we actually used redis as a backing store for pub sub (just as an example, we don't need to), sorted sets would satisfy the requirements of *exactly one*, predefined integer sort:

https://redis.io/topics/data-types#sorted-sets
https://redis.io/commands/zrangebyscore

This technique is explained in detail  here (just skip the java code in the middle):

https://www.infoworld.com/article/3212768/database/how-to-use-redis-for-real-time-stream-processing.html?page=2

I am not saying we have to use redis as a backing store (although I think it might be nice to support it as an optional backend), but more illustrating that if we simplify the requirements to string channels and search by `(exact channel, min height, max height)`, it becomes a much simpler problem to implement, as redis is a very simple k-v store.

I could also implement this sorted set query pattern on top of level db without much problem, and as anton has already implemented pubsub, that would cover the requirements.

ethanfrey:
btw, my approach for handling this directly on leveldb, which supports iteration, would be to define keys as `(channel string, block uint64, index uint32)` tuples, using uncompressed bigendian for the uints so their binary representation has the same lexographical order as an integer comparison. And prefixing string with the length to avoid overlap in searches between strings like "foo" and "foobar" (or another approach that acheives the same).

You could then calculate byte representations for `(channel, min block, 0)` and `(channel, max block, MAX_UINT32)` and run a level db iterator over that space. Super easy and efficient (edited)

melekes:
I like that idea, but the question is that enough for UI? https://allinbits.slack.com/archives/C564C4P1U/p1510910368000182?thread_ts=1510859626.000247&cid=C564C4P1U

melekes:
> What about all the uses of types.Result in the client package?

eh, a good thing to do probably. see
https://github.com/tendermint/abci/pull/130/commits/699c882eca65e0a2072aa8f238f2cb4846b29760

client/client.go
```diff
@@ -24,21 +24,21 @@ type Client interface {
...
 
 	FlushSync() error
-	EchoSync(msg string) (res types.Result)
-	InfoSync(types.RequestInfo) (resInfo types.ResponseInfo, err error)
-	SetOptionSync(key string, value string) (res types.Result)
-	DeliverTxSync(tx []byte) (res types.Result)
-	CheckTxSync(tx []byte) (res types.Result)
-	QuerySync(types.RequestQuery) (resQuery types.ResponseQuery, err error)
-	CommitSync() (res types.Result)
+	EchoSync(msg string) (*types.ResponseEcho, error)
```
melekes: these are breaking changes

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 0.8.0 (TBD)
+
+BREAKING CHANGES:
+ - [client] more consistent interface
```
ebuchman: Let's add some more details here about how

client/client.go
ebuchman: While we're at it, should we just make everything return (ResponseXxx, error) ? Seems like the best approach for future-proofing

test.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
ebuchman: thanks!

types/result.go
```diff
@@ -8,10 +8,12 @@ import (
 
 // Result is a common result object for ABCI calls.
 // CONTRACT: a zero Result is OK.
+// DEPRECATED: prefer raw types instead.
```
ebuchman: ++.

But we're making a breaking change. Let's just remove the file? I can't imagine it will be useful to anyone if all the interfaces are abandoning it :)

client/client.go
melekes: So, for example, App returns `ResponseDeliverTx{Code: CodeType_InternalError, Log: "bad"}` => abci client returns `nil, fmt.Errorf("%v: %s", r.Code, r.Log)` making interface more Golang style.. yeah, fine with me

ebuchman: Not quite. Client returning an error should indicate an internal error, like issue with connecting to the app. App errors should be recorded as ABCI code and should not show up as client error. Does that make sense ?

melekes: Ok, but we should make that clear in client docs

melekes:
Addressed all the comments.

types/base_app.go
```diff
@@ -7,36 +7,36 @@ func NewBaseApplication() *BaseApplication {
...
 
-func (app *BaseApplication) SetOption(key string, value string) (log string) {
+func (BaseApplication) SetOption(key string, value string) (log string) {
 	return ""
 }
 
-func (app *BaseApplication) DeliverTx(tx []byte) Result {
-	return NewResultOK(nil, "")
+func (BaseApplication) DeliverTx(tx []byte) ResponseDeliverTx {
+	return ResponseDeliverTx{Code: CodeType_OK}
```
melekes: btw, why DeliverTx/CheckTx/Commit return OK, but Query returns empty ResponseQuery{}?

ebuchman: LOL good eye. Should be effectively the same since OK is the zero value, but we should be consistent.

types/result.go
```diff
@@ -6,114 +6,47 @@ import (
 	"github.com/tendermint/go-wire/data"
 )
```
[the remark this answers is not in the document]
melekes: yes, in Tendermint

Makefile
```diff
@@ -36,12 +36,18 @@ dist:
...
+test: 
 	@ find . -path ./vendor -prune -o -name "*.sock" -exec rm {} \;
-	@ echo "==> Running unit tests"
+	@ echo "==> Running go test"
 	@ go test $(PACKAGES)
-	@ echo "==> Running integration tests (./tests)"
-	@ bash tests/test.sh
+
+test_race: 
+	@ find . -path ./vendor -prune -o -name "*.sock" -exec rm {} \;
```
melekes: this line is needed for `test_integrations`!

example/block_aware/block_aware_test.go
```diff
@@ -1,57 +0,0 @@
-package main
```
melekes: why remove this?

ebuchman: its not testing anything new. used to be when BlockAware was it's own interface. Now it's just part of Application.

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+BREAKING CHANGES:
+ - [client] all XxxSync methods now return (ResponseXxx, error)
+ - [types] Application: all methods now take RequestXxx and return (ResponseXxx, error).
+    - Except `CheckTx`/`DeliverTx`, which takes a `tx []byte` argument.
+    - Except `Commit`, which takes no arguments.
+ - [types] removed Result
+
+FEATURES:
+ - [types] added Tags field to ResponseDeliverTx
+ - [types] added Gas and Fee fields to ResponseCheckTx
```
melekes: want to add this to readme?

ebuchman: Yes thanks for reminder!

test.sh
```diff
@@ -0,0 +1,18 @@
...
+    if [ -f profile.out ]; then
+        cat profile.out >> coverage.txt
+        rm profile.out
+    fi
+done
+
+echo "==> Running integration tests (./tests)"
+find . -path ./vendor -prune -o -name "*.sock" -exec rm {} \;
+make install
+bash tests/test.sh
```
melekes: actually, we can call `make test_integrations` here

ebuchman: dont we want `test_integrations` to run the shell scripts too tho ?

melekes: oh, i see. ok

melekes:
Looks great 👍

ebuchman:
BOOM

PR #139, opened 2017-11-28
tests: sunset tmlibs/process.Process

odeke-em:
Updates https://github.com/tendermint/tmlibs/issues/81

No longer using tmlibs/process.Process as we deemed
it racy and would incur a maintenance cost yet not
used anywhere else but in these tests and not in actual
code.

### DIFF
--- a/tests/test_app/main.go
+++ b/tests/test_app/main.go
@@ -2,7 +2,10 @@ package main
 
 import (
 	"fmt"
+	"log"
 	"os"
+	"os/exec"
+	"time"
 
 	"github.com/tendermint/abci/types"
 )
@@ -20,6 +23,19 @@ func main() {
 	testCounter()
 }
 
+func ensureABCIIsUp(subCommand string, n int) error {
+	var err error
+	for i := 0; i < n; i++ {
+		cmd := exec.Command("bash", "-c", "abci-cli", subCommand)
+		_, err = cmd.CombinedOutput()
+		if err == nil {
+			break
+		}
+		<-time.After(500 * time.Second)
+	}
+	return err
+}
+
 func testCounter() {
 	abciApp := os.Getenv("ABCI_APP")
 	if abciApp == "" {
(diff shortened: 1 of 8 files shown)


tests/test_app/main.go
```diff
@@ -27,8 +30,17 @@ func testCounter() {
 	}
 
 	fmt.Printf("Running %s test with abci=%s\n", abciApp, abciType)
-	appProc := startApp(abciApp)
-	defer appProc.StopProcess(true)
+	cmd := exec.Command("bash", "-c", fmt.Sprintf("abci-cli %s", abciApp))
+	cmd.Stdout = os.Stdout
+	if err := cmd.Start(); err != nil {
+		log.Fatalf("starting %q err: %v", abciApp, err)
```
melekes: shouldn't we panic?

odeke-em: Same thing but here with more context.

tests/test_app/main.go
```diff
@@ -27,8 +30,17 @@ func testCounter() {
...
+	}
+	defer cmd.Wait()
+	defer cmd.Process.Kill()
+
+	// TODO: A better to wait for the app to start?
```
ebuchman: yes, we should be able to use `abci-cli echo` repeatedly until it works!

odeke-em: Would you like me to incorporate this change in right now? That sleep was a port over from the original code :)

ebuchman: Yeh I noticed. But I also fixed a bunch of these recently and non-deterministic tests are the worst, so would love if you fixed this!

odeke-em: Awesome, coming up.

odeke-em: C'est fini @ebuchman, merci beaucoup!

ebuchman:
@odeke-em see fix for ensureABCIIsUp

odeke-em:
Thank you very much. ahaha how did I forget to change `500 * time.Second` to `500 * time.Millisecond`??

The changes are quite comprehensive and nice, cool!

PR #142, opened 2017-11-30
testing automatic changes of byte array to data.Bytes

petabytestorage:
requested by bucky to be part of build pipeline

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 vendor
 .glide
+types/types.pb.go
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,11 @@ install_protoc:
 	go get github.com/golang/protobuf/protoc-gen-go
 
 protoc:
+	protoc --go_out=plugins=grpc:. types/*.proto
+	go run types/protoreplace/protoreplace.go
 	## On "error while loading shared libraries: libprotobuf.so.14: cannot open shared object file: No such file or directory"
 	##   ldconfig (may require sudo)
 	## https://stackoverflow.com/a/25518702
-	protoc --go_out=plugins=grpc:. types/*.proto
 
 install:
 	@ go install ./cmd/...
@@ -67,6 +68,7 @@ get_vendor_deps: ensure_tools
 
 metalinter:
 	protoc --lint_out=. types/*.proto
+	go run types/protoreplace/protoreplace.go
 	gometalinter --vendor --deadline=600s --enable-all --disable=lll ./...
 
 metalinter_test:
--- a/types/protoreplace/protoreplace.go
+++ b/types/protoreplace/protoreplace.go
@@ -1,0 +1,51 @@
+package main
+
+// +build ignore
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+//nolint: gas
+func main() {
+	bytePattern := regexp.MustCompile("[[][]]byte")
+	const oldPath = "types/types.pb.go"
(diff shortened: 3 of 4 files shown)


types/protoreplace.go
```diff
@@ -0,0 +1,51 @@
+package main
```
ebuchman: needs to go in its own package since this package is `types`

petabytestorage:
yes i will add those now

types/protoreplace/protoreplace.go
```diff
@@ -0,0 +1,53 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+// readLines reads a whole file into memory and returns a slice of its lines.
+func readLines(path string) ([]string, error) {
```
melekes: https://golang.org/pkg/io/ioutil/#ReadFile

```
content, err := ioutil.ReadFile(path)
	if err != nil {
		return nil, err
	}
bytes.Split(content, []byte("\n"))
```

petabytestorage: ty @melekes

Makefile
```diff
@@ -67,6 +68,7 @@ get_vendor_deps: ensure_tools
 
 metalinter:
 	protoc --lint_out=. types/*.proto
+	go run types/protoreplace/protoreplace.go
```
ebuchman: dont need this one!

types/types.pb.go
```diff
@@ -40,10 +39,10 @@ It has these top-level messages:
 	Validator
 	KVPair
 */
-//nolint: gas
```
ebuchman: oh ya, we also need to preserve this, since the generated code violates some rules of the linter.

maybe we can expand our tool to include the `nolint` statements we want?

types/types.pb.go
```diff
@@ -43,6 +44,7 @@ It has these top-level messages:
 //nolint: gas
```
ebuchman: this gets removed if I run `make protoc`. Believe it or not, this is the one we need (right above `package` statement), not the one at the very top!

.gitignore
```diff
@@ -1,2 +1,3 @@
 vendor
 .glide
+types/types.pb.go
```
melekes: are we removing it from Git? but abci won't compile without it, no?

petabytestorage:
adding to .gitignore does not remove it from git.

.gitignore
melekes: I know.. but if I make changes to it, I will not be able to commit them!

petabytestorage:
You can add and commit explicitly just not automatically without any command line option. -u and -f work

ebuchman:
Should we consider using gogo protobuf and their custom types as frey suggested?

https://github.com/gogo/protobuf

.gitignore
melekes: so what's the point of adding it to gitignore?

ebuchman: so we dont accidentally check in changes.

rudi-cilibrasi: To prevent automatic commit in git commit -a or plain git add. It’s a best practice for all auto generated files so that you don’t automatically commit every pointless variation to the repo like a time stamp or white space

melekes: hm.. ok

melekes:
If it will reduce the script size to 5 lines, yes. Otherwise, I would pass

ebuchman:
> If it will reduce the script size to 5 lines, yes. Otherwise, I would pass

My understanding is it would remove the need for the script all together!

ebuchman:
replaced by https://github.com/tendermint/abci/pull/145

PR #145, opened 2017-11-30
Types makeover

ebuchman:
- remove CodeType
- move to using `gogo/protobuf` instead of `google/protobuf`
- use `customtype` from `gogo/protobuf` so we can replace all `[]byte` with `data.Bytes`
- remove `ResultQuery`


ebuchman:
TODO: bring in rudis script

ebuchman:
TODO: make corresponding change to tendermint tutorials that print messages for the code types

client/grpc_client.go
```diff
@@ -68,6 +69,7 @@ RETRY_LOOP:
 			if err == nil {
 				break ENSURE_CONNECTED
 			}
+			cli.Logger.Info("Echo failed", "err", err)
```
melekes: `Error`?

example/code/code.go
```diff
@@ -0,0 +1,9 @@
+package code
+
+// Return codes for the examples
+const (
+	CodeTypeOK            uint32 = 0
```
melekes: you can use `iota` https://dlintw.github.io/gobyexample/public/constants-and-iota.html

tests/test_cli/ex2.abci.out
```diff
@@ -11,14 +11,14 @@
 -> code: OK
 
 > check_tx 0x00
--> code: BadNonce
+-> code: 2
```
melekes: don't we want to write some `humanCode` helper for that?

example/code/code.go
ebuchman: I don't like iota for things that are cross-language standards because then you have to count to figure out which one is which. Granted it's ok here since there's so few and it's easier to maintain.

tests/test_cli/ex2.abci.out
ebuchman: We do, but it's not really easy to do in general since the codes will be the responsibility of the app. We have an example code now in `example/code` but the `abci-cli` is more general than that so it shouldn't print `BadNonce` when it might be talking to an app for which `2` means `BadGong` or whatever.

ebuchman: We should keep thinking about how to solve this but I don't have a good solution yet

melekes: got it. I was talking more about providing an example of such wrapper (possibly inside an app) to devs.

ebuchman: Opened an issue: https://github.com/tendermint/abci/issues/146

ebuchman: > got it. I was talking more about providing an example of such wrapper (possibly inside an app) to devs.
 
That'd be nice but I still don't see how we'd be able to use it here since `abci-cli` is too general purpose (it doesn't load an app in or anything). Good think we have `log`

example/code/code.go
odeke-em: True true

example/dummy/dummy_test.go
```diff
@@ -36,7 +39,7 @@ func testDummy(t *testing.T, app types.Application, tx []byte, key, value string
 		Data:  []byte(key),
 		Prove: true,
 	})
-	require.Equal(t, types.CodeType_OK, resQuery.Code)
+	require.EqualValues(t, code.CodeTypeOK, resQuery.Code)
```
odeke-em: Minor nit: above we use `Equal`, then here `EqualValues`, but that's alright :)

types/protoreplace/protoreplace.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+		if gotPackageLine {
+			fmt.Fprintf(outFile, "%s\n", "//nolint: gas")
+		}
+		fmt.Fprintf(outFile, "%s\n", line)
+	}
+	outFile.Close()
+	os.Remove(oldPath)
+	os.Rename(tmpPath, oldPath)
+	exec.Command("goimports", "-w", oldPath)
```
odeke-em: It'd be nice to have error handling for this as perhaps new developers might not have `goimports` setup.

PR #147, opened 2017-12-01
Int64

ebuchman:


odeke-em:
@ebuchman this change does more than just the int64 usage though, it also uses a different protobuf repo, could we perhaps separate those or at least mention it in the commit info? Sorry to sound like a stickler but I believe it will help things in general to isolate such changes because the protobuf provider change is a big one :)

ebuchman:
this doesn't change the provider. that was done in https://github.com/tendermint/abci/pull/145. Would love another pair of eyes on that one too if you have the chance :)

odeke-em:
Ooh I see, thanks for the reference, let me take a look at it.

odeke-em:
I've given #145 some eyes, looks great, nothing jumped out the page, except just the suggestion to handle the error after invoking `goimports`

types/types.proto
```diff
@@ -178,13 +178,13 @@ message BlockID {
 }
 
 message PartSetHeader {
-  uint64 total = 1;
+  int64 total = 1;
```
melekes: Are we subtracting total somewhere? I thought we just set it once

types/types.proto
```diff
@@ -162,9 +162,9 @@ message ResponseEndBlock{
 
 message Header {
   string chain_id = 1;
-  uint64 height = 2;
-  uint64 time = 3;
-  uint64 num_txs = 4;
+  int64 height = 2;
+  int64 time = 3;
+  int64 num_txs = 4;
```
melekes: same here. or we decided to completely get rid of uint64?

ebuchman: You're probably right, but I think best to just be consistent, no?

PR #156, opened 2017-12-10
cmd/abci-cli: use a single connection per session

### LINKED ISSUE #133: abci-cli console/batch shouldn't open a new connection with each command
See https://github.com/tendermint/abci/issues/125

odeke-em:
Use the single client connection at startup time
for sending over commands instead of shelling out
for every command.
This code fixes the regression from
https://github.com/tendermint/abci/pull/117
which instead used "os/exec".Command with:
    "abci-cli <the_command> [args...]"

The purpose of this code is to restore us
back to the state after cobra replace urlfave/cli.
There is still a bit of work to implement Batch
itself, but that should be simpler as a focused
command.

Fixes #133

### DIFF
--- a/cmd/abci-cli/abci-cli.go
+++ b/cmd/abci-cli/abci-cli.go
@@ -359,14 +359,11 @@ func cmdBatch(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
-		if len(line) > 0 { // prevents introduction of extra space leading to argument parse errors
-			args = strings.Split(string(line), " ")
-		}
-
-		if err := muxOnCommands(cmd, args); err != nil {
+		cmdArgs := persistentArgs(line)
+		if err := muxOnCommands(cmd, cmdArgs); err != nil {
 			return err
 		}
-		fmt.Println("")
+		fmt.Println()
 	}
 	return nil
 }
@@ -397,22 +394,14 @@ var (
 )
 
 func muxOnCommands(cmd *cobra.Command, pArgs []string) error {
-
-	if len(pArgs) < 1 && cmd.Use != "batch" {
+	if len(pArgs) < 2 {
 		return errBadPersistentArgs
 	}
-
-	subCommand, actualArgs := pArgs[0], pArgs[1:]
-
-	//if cmd.Use == "batch" {
-	//	cmd.Use = subCommand
-	//}
-
+	subCommand, actualArgs := pArgs[1], pArgs[2:]
 
 	switch strings.ToLower(subCommand) {
 	case "batch":
-		fmt.Println("WTF")
-		return nil
+		return muxOnCommands(cmd, actualArgs)
 	case "check_tx":
 		return cmdCheckTx(cmd, actualArgs)
 	case "commit":
@@ -437,6 +426,9 @@ func cmdUnimplemented(cmd *cobra.Command, args []string) error {
 	msg := "unimplemented command"
 	if err := cmd.Help(); err != nil {
 		msg = err.Error()
+	}
+	if len(args) > 0 {
+		msg += fmt.Sprintf(" args: [%s]", strings.Join(args, " "))
 	}
 	printResponse(cmd, args, response{
 		Code: codeBad,
--- a/cmd/abci-cli/abci-cli.go
+++ b/cmd/abci-cli/abci-cli.go
@@ -168,7 +168,7 @@ var consoleCmd = &cobra.Command{
 	Short:     "Start an interactive abci console for multiple commands",
 	Long:      "",
 	Args:      cobra.ExactArgs(0),
-	ValidArgs: []string{"batch", "echo", "info", "set_option", "deliver_tx", "check_tx", "commit", "query"},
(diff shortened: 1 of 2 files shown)


zramsay:
dope! Looks like I was on the right path originally :)

zramsay:
i've tested the `abci-cli console` and it works great :+1:

cmd/abci-cli/abci-cli.go
```diff
@@ -394,14 +397,22 @@ var (
...
 
 func muxOnCommands(cmd *cobra.Command, pArgs []string) error {
-	if len(pArgs) < 2 {
+
+	if len(pArgs) < 1 && cmd.Use != "batch" {
 		return errBadPersistentArgs
 	}
 
-	subCommand, actualArgs := pArgs[1], pArgs[2:]
+	subCommand, actualArgs := pArgs[0], pArgs[1:]
```
odeke-em: @zramsay how come we changed this? That'll make it fail for all the other commands as they use persistentArgs. 
For now am going to revert just this slicing part, and instead make persistentArgs for batch too, how does that sound?

odeke-em:
@zramsay, please take a look, I've implemented a version of batch to take in commands from stdin,
thanks to an internal consultation from @ebuchman. I think TBH that "batch" for console doesn't make sense unless some hackasaurus can redirect stdin after we've started reading from it. However, this PR returns us to the parity we had before so the change about "batch" for console can come later on.

zramsay:
https://github.com/tendermint/abci/blob/use-single-connection-in-console/tests/test_cli/test.sh#L20
^ this line fails & gives:
```
You broke the tutorial
Got:
Run a batch of abci commands against an application

Usage:
  abci-cli batch [flags]

Flags:
  -h, --help   help for batch

Global Flags:
      --abci string        Either socket or grpc (default "socket")
      --address string     Address of application socket (default "tcp://0.0.0.0:46658")
      --log_level string   Set the logger level (default "debug")
  -v, --verbose            Print the command and results as if it were a console session
> batch abci-cli --log_level=error --verbose echo hello
-> code: 10
-> log: unimplemented command args: [abci-cli --log_level=error --verbose echo hello]
```

cmd/abci-cli/abci-cli.go
```diff
@@ -382,18 +380,70 @@ func cmdConsole(cmd *cobra.Command, args []string) error {
...
 		}
-		fmt.Println(string(out))
 	}
 	return nil
 }
 
+var (
+	errBadPersistentArgs = errors.New("expecting persistent args of the form: abci-cli [command] <...>")
+
+	errUnimplemented = errors.New("unimplemented")
```
melekes: unused?

cmd/abci-cli/abci-cli.go
```diff
@@ -382,18 +380,70 @@ func cmdConsole(cmd *cobra.Command, args []string) error {
...
+		if err := muxOnCommands(cmd, pArgs); err != nil {
 			return err
 		}
-		fmt.Println(string(out))
 	}
 	return nil
 }
 
+var (
+	errBadPersistentArgs = errors.New("expecting persistent args of the form: abci-cli [command] <...>")
```
melekes: why a var if it only used in one place?

melekes:
What's up with this PR?

cmd/abci-cli/abci-cli.go
odeke-em: Great question, that's a vestige of the former code. I'll remove it.

cmd/abci-cli/abci-cli.go
odeke-em: Yeah before we hadn't implemented a couple of funcs, I'll remove it.

odeke-em:
I was hoping that the initial purpose of this PR would be to return the console back to its original state of using a single connection, the actual command logic would then be implemented otherwise I am currently spending a lot of time on this PR than I'd ideally have wanted to. Could someone help me out with finishing that part or at least step me through these commands? Also what was the impetus for us to switch CLI parsers?

odeke-em:
I was hoping that the initial purpose of this PR would be to return the console back to its original state of using a single connection, the actual command logic would then be implemented as a follow-up. I mention this because I am currently spending a lot of time on this PR than I'd ideally have wanted to. Could someone help me out with finishing that part or at least step me through these commands? Also what was the impetus for us to switch CLI parsers?

ebuchman:
This looks awesome but it's broken because it can't handle flags.

ie. if I start the console with `abci-cli --log_level=error --verbose console` I can't run any commands because it thinks the flags are the commands. So we may have to be more intelligent about parsing out the command ... @odeke-em @zramsay

ebuchman:
This is why the test is failing:

```
-> log: unimplemented command args: [abci-cli --log_level=error --verbose echo hello]
```

ebuchman:
I got it

odeke-em:
Thank you @ebuchman -- mah dawg #DenzelWashingtonVoice

PR #157, opened 2017-12-10
enable linter & make deterministic

zramsay:


Makefile
```diff
@@ -71,11 +71,11 @@ get_vendor_deps: ensure_tools
 
 metalinter:
 	protoc $(INCLUDE) --lint_out=. types/*.proto
-	gometalinter --vendor --deadline=600s --enable-all --disable=lll ./...
+	gometalinter.v2 --vendor --deadline=600s --enable-all --disable=lll ./...
```
melekes: we don't need the same `$(GOPATH)/bin/` here?

zramsay: yea, not sure why. Suspect it has to do with the circle setup

Makefile
```diff
@@ -71,11 +71,11 @@ get_vendor_deps: ensure_tools
...
 
 metalinter_test:
-	protoc $(INCLUDE) --lint_out=. types/*.proto
-	gometalinter --vendor --deadline=600s --disable-all  \
+	#protoc $(INCLUDE) --lint_out=. types/*.proto
```
melekes: either remove or uncomment

Makefile
melekes: ok

PR #159, opened 2017-12-12
types: compile type assertions to avoid sneaky runtime surprises

odeke-em:
Ensure that the types in result.go implement both
json.Marshaler and json.Unmarshaler
and thus avoid any accidental deletions of their respective
methods which would then cause surprises at runtime.

### DIFF
--- a/types/result.go
+++ b/types/result.go
@@ -2,6 +2,7 @@ package types
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


odeke-em:
Thank you @petabytestorage for the quick review and approval!

PR #160, opened 2017-12-12
Short-circuit assertions in test command

KrzysiekJ:
Refs #76.

The implementation relies on the fact that subsequent server calls don’t share any state. If it was not the case, the code would need to be more complicated, more monadic (vide [“Monads for Go programmers”](https://awalterschulze.github.io/blog/post/monads-for-goprogrammers/)).

### DIFF
--- a/cmd/abci-cli/abci-cli.go
+++ b/cmd/abci-cli/abci-cli.go
@@ -300,49 +300,37 @@ func persistentArgs(line []byte) []string {
 
 //--------------------------------------------------------------------------------
 
-func or(err1 error, err2 error) error {
-	if err1 == nil {
-		return err2
+func compose(fs []func() error) error {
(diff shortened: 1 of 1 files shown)


melekes:
I'd vouch against this PR. It makes code more complicated. Don't get me wrong. I love functional programming (I was an Erlang dev before), but trying to bring FP concepts to the language that was not designed for it is just feels wrong. Again, this is my personal imho.

KrzysiekJ:
How to short-circuit these assertions without functional means and without staircased ifs then? (Assuming we agree that they should be short-circuited).

ebuchman:
This is nice. @melekes how is this worse than before? Seems cleaner to me

melekes:
Sorry, I was in anti-functional mood. Indeed, looks better

PR #161, opened 2017-12-13
all: fix vet issues with build tags, formatting

odeke-em:
* Build tags need to come before the package name
and have at least a blank line, between them and
the package, please see
  https://golang.org/pkg/go/build/#hdr-Build_Constraints
* fmt.Println doesn't take formatting verbs
* Fix a missing formatting argument to fmt.Printf

### DIFF
--- a/tests/server/client.go
+++ b/tests/server/client.go
@@ -21,7 +21,7 @@ func InitChain(client abcicli.Client) error {
 	}
 	_, err := client.InitChainSync(types.RequestInitChain{Validators: vals})
 	if err != nil {
-		fmt.Println("Failed test: InitChain - %v", err)
+		fmt.Printf("Failed test: InitChain - %v\n", err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


tests/server/client.go
```diff
@@ -46,7 +46,7 @@ func Commit(client abcicli.Client, hashExp []byte) error {
 	_, data := res.Code, res.Data
 	if err != nil {
 		fmt.Println("Failed test: Commit")
-		fmt.Printf("committing %v\nlog: %v", res.GetLog())
+		fmt.Printf("committing %v\nlog: %v", data, res.GetLog())
```
melekes: can you add "\n" to the end here too?

melekes: done

odeke-em:
Thanks for that change to add "\n" to the touched lines @melekes. @ebuchman PTAL.

PR #164, opened 2017-12-17
protoc: "//nolint: gas" directive after pb generation

### LINKED ISSUE #138: types/types.pb.go: needs linting/`nolint` tag ...
types.pb.go doesn't satisfy the `gas` linter since it doesn't check some errors.

we need to automatically add `//nolint: gas` above the `package` statement after running `protoc`.

At least until we get the Go protobuf team to lint their generated code.

@odeke-em :) ?

odeke-em:
Fixes #138

Since we can't add package directives through the protoc
compiler, yet we need to "//nolint: gas" the Go generated
protobuf file, added a script whose purpose is to
go find the "package (\w+)$" declaration after go fmt
was run by protoc.

The competing solutions were more complex and can be
examined by visiting
https://github.com/tendermint/abci/issues/138#issuecomment-352226217

### DIFF
--- a/scripts/ducttape_nolint_protoc.sh
+++ b/scripts/ducttape_nolint_protoc.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+grep -Rn 'package \(\w\+\)$' types/*.pb.go | cut -d':' -f1 | while read F;do backup="$F.bak";sed "s/package \([a-zA-Z]\{1,\}\)$/\/\/nolint: gas<NEWLINE_HEREXXXX>&/1" "$F" | awk -F"<NEWLINE_HEREXXXX>" '{ if (NF >= 2) { print $1"\n"$2 } else { print $1 } }' > "$backup";mv "$backup" "$F";done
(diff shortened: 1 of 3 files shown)


scripts/ducttape_nolint_protoc.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+grep -Rn 'package \(\w\+\)$' types/*.pb.go | cut -d':' -f1 | while read F;do backup="$F.bak";sed "s/package \([a-zA-Z]\{1,\}\)$/\/\/nolint: gas<NEWLINE_HEREXXXX>&/1" "$F" | awk -F"<NEWLINE_HEREXXXX>" '{ if (NF >= 2) { print $1"\n"$2 } else { print $1 } }' > "$backup";mv "$backup" "$F";done
```
melekes: too complicated... maybe we should just do something like `sed  "/package types/ { N; s/\/\/nolint: gas\n/package types\n&/ }" types/types.pb.go`

melekes: I will do later

melekes:
@odeke-em simplified script. please comment if I am missing something

odeke-em:
Thank you @melekes! So on my Mac, the script fails with
```shell
protoc -I=. -I=/Users/emmanuelodeke/go/src -I=/Users/emmanuelodeke/go/src/github.com/gogo/protobuf/protobuf --gogo_out=plugins=grpc:. --lint_out=. types/*.proto
# Since protoc can't "//nolint: gas" our packages, we need to do it manually.
sed -i "/package types/ { N; s/package types\n/\/\/nolint: gas\n&/ }" types/*.pb.go
sed: 1: "types/types.pb.go": undefined label 'ypes/types.pb.go'
make: *** [protoc] Error 1
```

odeke-em:
@melekes I actually think that the original form wasn't any less complicated than the new one plus it was a 1 liner :)

PR #184, opened 2018-02-02
update for go-wire sdk2

melekes:


adrianbrink:
I think we want to upgrade all the dependencies to the `develop` branches. 

This can be the replacement: https://github.com/tendermint/abci/pull/185

PR #193, opened 2018-02-17
types: RequestInitChain.AppStateBytes

ebuchman:


ethanfrey:
AppStateBytes are basically the section that used to be called AppOptions in the genesis file?

Would be good to add a comment or two to explain where this comes from (serialized is a bit vague)

PR #208, opened 2018-03-05
update spec formatting for docs

zramsay:
this file is pulled in by tendermint's doc building


zramsay:
merging  because fixing tendermint docs is blocking on this

PR #209, opened 2018-03-05
Move the tmlibs dependency to develop

zmanian:
Working on fixing transitive dependencies so https://github.com/cosmos/cosmos-sdk/pull/546 works


zramsay:
looks legit

PR #213, opened 2018-03-12
types: compile time assert to, and document sort.Interface

### LINKED ISSUE #212: Validators: remove verbose sort.Interface methods in favour of sort.Slice
Raised offline by an auditor, and I totally see where they are coming from, the code for Validators implements that sort.Interface interface but doesn't conform to it with a compile type assertion to convey the message to auditors and other folks who are auditing method by method.
https://github.com/tendermint/abci/blob/68592f4d8ee34e97db94b7a7976b1309efdb7eb9/types/util.go#L11-L26

Tendermint now runs on Go1.8+ hence we can can now simply invoke
```go
sort.Slice(validators, func(i, j int) bool {
   return bytes.Compare(v[i].PubKey, v[j].PubKey) <= 0
})
```
It would be nicer to get rid of the older verbose code and keep things less ambiguous. This will help reduce the code base, keep it modern, reduce the API surface but even better not trigger redlights for anyone inspecting method by method.

odeke-em:
Fixes #212

Declare the purpose of the Less, Len, Swap methods
so that readers can see why they are defined.

Raised by an auditor in their report, as it looked like a security
concern but actually sort.Interface requires those methods implemented.

### DIFF
--- a/types/util.go
+++ b/types/util.go
@@ -3,12 +3,21 @@ package types
 import (
 	"bytes"
 	"encoding/json"
+	"sort"
 )
 
 //------------------------------------------------------------------------------
 
 // Validators is a list of validators that implements the Sort interface
 type Validators []Validator
+
+var _ sort.Interface = (Validators)(nil)
+
+// All these methods for Validators:
+//    Len, Less and Swap
+// are for Validators to implement sort.Interface
+// which will be used by the sort package.
+// See Issue https://github.com/tendermint/abci/issues/212
 
 func (v Validators) Len() int {
 	return len(v)


odeke-em:
@xla great question, so initially I was hoping to do the sorting inline but it's used in about 4 spots so didn't want to copy all the content allover.

odeke-em:
I believe the interface conformance will solve the problem for the auditor not knowing why those methods are how they look but also prevent duplication.

ebuchman:
Can't we just define a SortValidators function that uses the new sort.Slice and call that in the 4 places ? Seems like the best approach

PR #220, opened 2018-03-27
Add InitialValSet and AppHash to ResponseInitChain

adrianbrink:

PR #222, opened 2018-03-27
WIP: Add genesis information to RequestInitChain

cwgoes:
Ref https://github.com/tendermint/abci/issues/216
Ref https://github.com/cosmos/cosmos-sdk/issues/661

What other genesis information do we want in `RequestInitChain`?


cwgoes:
Duplicate of https://github.com/tendermint/abci/pull/220

Issue #232, opened 2018-05-10
Given a global flag to configure the application listen address, why non-global flags needed?


maxim-levy:
There is a global flag for the address defined as:

> RootCmd.PersistentFlags().StringVarP(&flagAddress, "address", "", "tcp://0.0.0.0:46658", "address of application socket")

Then, what is the meaning for command-specific `addr` flags, like this one:

https://github.com/tendermint/abci/blob/78a8905690ef54f9d57e3b2b0ee7ad3a04ef3f1f/cmd/abci-cli/abci-cli.go#L138

ebuchman:
Seems like a redundancy - there should just be the global address.

Would you be interested in making a PR to clean this up ?

maxim-levy:
Let me try

zramsay:
merged to develop

PR #238, opened 2018-05-16
A redundant "listen address" flag removed

maxim-levy:
To address issue #232


zramsay:
merging to `max/fix/l_addr` branch for the CI to run

PR #243, opened 2018-05-17
Add evidence type and block timestamp to abci.Evidence

cwgoes:
ABCI applications implementing any kind of time-aware slashing scheme (e.g. an unbonding period) need the timestamp of the block which was double-signed.


ebuchman:
I think we'll need to iterate on this some more but fine for now

PR #245, opened 2018-05-20
Request/ResponseValidator, update Header

ebuchman:
# Remark on Philosophy

The goal of ABCI is to be maximally flexible across languages, applications, consensus engines, and to be easy for developers to use. Thus we need to strike the right balance between what gets included in protobuf and what is left out, and also which Tendermint specific data structures get included or modified.

I think the proposal here better achieves these goals.

## PubKey

A pubkey has a type (string) and some data (bytes). It's the simplest least contentious thing imaginable. Let's start with this. For now, we will have `type="ed25519"` and `data=<32-byte pubkey>`

## Validator

Replace with two new types: `RequestValidator` and `ResponseValidator`. The former is sent in requests from Tendermint, and includes all info about the validator (pubkey, index, address, voting power). This way its the full amount of information available to make things as easy as possible for devs - for instance they can use the address as a key, but don't need to know how to compute it. The later is returned in responses and contains just the required info - the pubkey and the power.

## Header

I'm not sure all those values in the header are really that useful, so I gutted all hashes except the last block hash and the app hash, which are hashes the app actually deals in.


types/types.proto
```diff
@@ -214,38 +215,50 @@ message BlockGossip {
...
   int64 time = 3;
+
+  // txs
   int32 num_txs = 4;
-  BlockID last_block_id = 5 [(gogoproto.customname)="LastBlockID", (gogoproto.nullable)=false];
-  bytes last_commit_hash = 6;
-  bytes data_hash = 7;
-  bytes validators_hash = 8;
-  bytes app_hash = 9;
+  int64 total_txs = 5;
```
melekes: what's the difference between num_txs and total_txs?

types/types.proto
```diff
@@ -214,38 +215,50 @@ message BlockGossip {
...
+  PubKey pub_key = 1;
+  int64 power = 2;
 }
 
-message PartSetHeader {
-  int32 total = 1;
-  bytes hash = 2;
+// Validator for use in requests
+message RequestValidator {
+  bytes address = 1;
```
melekes: why do we send address if it can be derived from pubkey?

melekes: so the app dev. does not need to calculate it herself?

types/types.proto
b00f: I think num_tx is the number of transaction in this block, but total_tx is total transaction since genesis time.

types/types.proto
b00f: I also have this question, why we need publicKey and address at the same time?
Also I am wondering if it is possible to include public key in the vote or commits (so each vote's can be checked without looking at the validator's public_key in the validator set) and in the genesis document validators list by their address (not public key). It's more easy to manage the genesis doc by address rather than pub_key.

I hope I could explain my idea...

types/types.proto
melekes: right. I forgot. what happens when total_txs overflows? @ebuchman

b00f: It should be end of the world! 2 in power of 64 transactions!

types/types.proto
ebuchman: We don't need both, but this gives maximum flexibility to the app. I'd be happy to just pass the pubkeys in too, without the address. What about the index ? 

> Also I am wondering if it is possible to include public key in the vote or commits (so each vote's can be checked without looking at the validator's public_key in the validator set) 

Would be preferable not to make the votes bigger like this. Do you need this for some reason ? You shouldnt need to check votes directly in the app (well, other than for IBC).

> and in the genesis document validators list by their address (not public key). It's more easy to manage the genesis doc by address rather than pub_key.

Here we need pubkey so that we know what the pubkey is for verification since we dont want to include them in the votes

ebuchman: > We don't need both, but this gives maximum flexibility to the app. I'd be happy to just pass the pubkeys in too, without the address. What about the index ?

Actually the benefit of the address is apps can use it without worrying about how to compute it ... maybe that is helpful. Why not give all the data?

b00f: I agree, the application layer doesn't need to check or validate the votes.  However, it seems better if each vote has enough information (by itself) to be validated without having to look into the validator set via the index. I would say that the Index is not really relevant or helpful to include in the vote data because the index can vary from round to round.

Let me explain what I'm thinking. 

Recently I am working on a project based on Tendermint and Hyperledger Burrow. We are developing a Proof Of Stake (not Delegated Proof Of Stake) consensus model where there is a pool of validators and in each round members of this pool can select themselves by Verifiable Random /function (VRF) to enter the Tendermint validator set. In this model, a validator with more stake has more chance to be chosen from the pool, however the validator power in Tendermint remains uniform and decoupled from stake so that we can achieve a maximally democratic vote model; it means every node has equal vote power.

On the application side, in the IVAL tree, we store the information of each validator by their address. With a bond transaction, a new validator can join the pool and with an unbond transaction a validator can leave the pool.

https://github.com/finterra/finterra-burrow

In Burrow (which is the primary base code we are using for the application layer over Tendermint) each account holds the address and the public_key as a part of this account information. I believe the application layer only needs the public_key for validating transactions (i.e., validating the signature and validating the txn sender address). However, beacuse pub_keys are always included in the transactions themselves, in the whole eco-system the application layer can manage everything using the address alone (for both accounts and validators). In this case, the application layer only needs to update the validator set information using validators' addresses.

My thought is why not have all validator information (including their voting status and history) queryable based on their address as a key in both Tendermint and the application layer (not the public_key) 

If public_key data is included in every vote (or transaction) then we don't need to pass the public_key for validating in Tendermint, does that seem right???

Additionally, I have another related but separate suggestion for addresses. 

In go-crypto (which is a lovely project) it will be helpful if we can add checksum to the address object. This way addresses can be validated by themselves (like Bitcoin addresses), which would avoid much user error by avoiding sends to blackhole addresses ;) :
https://bitcoin.stackexchange.com/questions/32353/how-do-i-check-the-checksum-of-a-bitcoin-address

ebuchman: Thank you for this thoughtful response!

First, about checksums, we have been working on a standard called bech32cosmos: https://github.com/tendermint/go-crypto/pull/66

It should get merged this week!!

As for pubkeys in votes. I understand that it would nice if the votes included the pubkey so its easy to verify them without access to any other information. However:

1) you still need to check that the pubkey is actually part of the validator set, which means you need access to the validator set anyways, which means you don't really gain anything by including the public key in the vote

2) gossipping votes is the most time consuming part of the algorithm. anything we add to votes needs to be very well justified otherwise it may just unecesarily slow down the consensus

3) we want to be able to support BLS signature aggregation in the future which requires all votes be identical except the signature - if we include the pubkey, it would ruin this. BLS will allows Tendermint to become MUCH more performant and scalable, and can also provide a built in source of randomness!

Now, I'm still trying to understand why you're asking for this. Are you saying you don't want to have to pass the PubKey back in EndBlock, and instead just use the address ?

types/types.proto
```diff
@@ -61,8 +61,9 @@ message RequestQuery {
 message RequestBeginBlock {
   bytes hash = 1;
   Header header = 2 [(gogoproto.nullable)=false];
-  repeated bytes absent_validators = 3;
-  repeated Evidence byzantine_validators = 4 [(gogoproto.nullable)=false];
+  repeated RequestValidator present_validators = 3;
```
ebuchman: we can just pass one list of validators here, where each has a bool that says whether they signed or not. and we don't need the index since they'll be in order, but might as well still have pubkey, address, power

types/types.proto
b00f: Thanks @ebuchman. 

My thought is more based on this idea that we need to capsulate the public-key and signature and not exposing to them. 

In my point of view, it will be better if the public key is not exported publicly at all. Public keys transfer among the signature and anytime a vote is needed to be verified, it can be done without exposing the public key. The address derived from the public key might later be checked whether it exists in validator set or not. In this case public-key won't be passed between objects. It helps to maintain the code much more better.

For supporting BLS signature aggregation, the public-key can be sited next to the signature. So in this case other part of the vote will be identical all over the network and each vote can be verified easily without looking at the validator set. 

This idea came to my mind when I implemented validator-pool in our project. In the application layer anyone can be a validator by bonding some tokens. A bonding transaction should be same as send transaction which transfer some token between accounts. However, to make a bond-tx (and also unbod-tx) more unify in the application layer, we should avoid public key as a part of transaction. Please take a look at the implementation of bond-tx here:

https://github.com/finterra/finterra-burrow/blob/master/txs/bond_tx.go

As you can see here, it is almost same as send-tx (sending x tokens from addr_1 to addr_2), however because tendermint needs to have public key in EndBlock, it's needed to pass the public-key as well the address which is I believe is unnecessary here.

I hope I could explain my idea here.

ebuchman: I think this is a really valuable idea. Thanks for taking the time to communicate this request. I've opened an issue on the Tendermint repo to consider it: https://github.com/tendermint/tendermint/issues/1625

Thanks!

Issue #179, opened 2018-01-12
types.proto missing import


mdyring:
The import github.com/gogo/protobuf/gogoproto/gogo.proto doesn't exist which causes protoc to complain.

In addition some concerns from Matrix chat that would be nice to address:

srmo: Could someone elaborate ob the new gogo types in the abci protobuf types definition? It became pretty difficult to port this to non go languages. Are those new types required? What happens if we leave them out? What are you using them for?

mdyring: We should probably open an issue for it, I haven't tested yet on C++.. But are you seeing build errors?

srmo: @mdyring i had to do crazy workarounds to generate the class files in java which worked but i had no idea why. Someone opened a PR with the gogoproto stuff removed, which has no side effects in simple test cases but this isn’t really providing confidence. A issue seems the way to go.

sbellem:
We ran into this issue meanwhile trying to upgrade the Python ABCI project, and so far I have not found a straightforward solution. 

The CockroachDB project had a similar [issue](https://github.com/cockroachdb/cockroach/issues/393) and seemingly resolved it by making gogo-free proto files.

In another related [issue](https://github.com/prometheus/prometheus/issues/3270#issuecomment-335821242), on the Prometheus project, performance is given as a key reason:

> With the state of things, not using gogoproto is simply not an option from a performance and code sanity perspective.

it is suggested to remove the language specific annotations:

> The only reasonable answer right now is to copy the file, strip all annotations and potentially replace them with annotations for the language of your choice. It's an awful user experience. But that's for the protobuf/grpc community to figure out.

srmo:
thanks @sbellem for the links to similar issues. I'll look into the last two points. We have the same problem with the java port.

sbellem:
> thanks @sbellem for the links to similar issues. I'll look into the last two points. We have the same problem with the java port.

Thanks to you too! I saw the issue on `jabci` and it was useful, as it at least confirmed I was not alone! :)

srmo:
:) yep, that's always a good feeling

johndpope:
FYI  this maybe of some use. https://github.com/tendermint/abci/tree/master/types/protoreplace This I understand predates gogo/profobuffer. Perhaps need to just trawl proto file history to find a working file.

sbellem:
> FYI this maybe of some use. https://github.com/tendermint/abci/tree/master/types/protoreplace

Thanks @johndpope.

ebuchman:
[...]

PR #247, opened 2018-05-24
Bucky/no gogo

ebuchman:
Depends on https://github.com/tendermint/tmlibs/pull/214

Remove gogoproto!

Refs https://github.com/tendermint/abci/issues/179


types/result.go
```diff
@@ -1,121 +1,170 @@
...
-	return []byte(s), err
+type ResultCheckTx struct {
+	Code      uint32       `json:"code,omitempty"`
+	Data      []byte       `json:"data,omitempty"`
+	Log       string       `json:"log,omitempty"`
+	Info      string       `json:"info,omitempty"`
+	GasWanted int64        `json:"gas_wanted,omitempty"`
+	GasUsed   int64        `json:"gas_used,omitempty"`
+	Tags      []cmn.KVPair `json:"tags,omitempty"`
+	Fee       cmn.KI64Pair `json:"fee"`
```
melekes: how do you choose whenever to add `omitempty` or not?

melekes: it correlates with `(gogoproto.nullable)=false`, right?

ebuchman: oh good point. I didn't choose. I just left it how it was. yeh i think it does correlate

PR #252, opened 2018-05-30
Change evidence type to enum

cwgoes:
Ref https://github.com/cosmos/cosmos-sdk/pull/1011#discussion_r191904300


cwgoes:
Probably we should update Tendermint/SDK to latest ABCI develop first.

ebuchman:
I think a string would actually be best, like in PubKey now.

A pubkey assumes data is just bytes. For Evidence we assume data includes a pubkey, height, and time, but otherwise the type doesnt change the kind of information included.

Using a string could allow the type to take hierarchical form - "equivocation/proposal" or "invalid/proposal/signature".

If we need to provide even more information, we can later add a `bytes data` field and let the type string also inform us how it's encoded ...

cwgoes:
Can we include some constant types in this repository then, or find some other way to import a common set in Tendermint and the SDK?

ebuchman:
Yep let's add the type strings to tendermint/types/evidence.go

cwgoes:
Great, closing in favor of https://github.com/tendermint/tendermint/pull/1652

PR #253, opened 2018-05-31
Switch port 46658 to 26658

ValarDragon:
This moves usage of port 46658 to 26658. This is done to address cosmos/cosmos-sdk#932. 
Essentially the problem was the default ports used by tendermint and the sdk are in the ephemeral port range on Linux.


cwgoes:
Looks like this was branched from `develop` but is targeted on `master`?

ValarDragon:
Sorry, didn't mean to merge this. (I did commands on the command line wrong), but it was apparently okay to merge.

PR #254, opened 2018-05-31
include genesis file structure

ebuchman:


types/types.proto
```diff
@@ -47,8 +47,11 @@ message RequestSetOption {
 }
 
 message RequestInitChain {
-  repeated Validator validators = 1 [(gogoproto.nullable)=false];
-  bytes genesis_bytes = 2;
+  int64 time = 1;
```
adrianbrink: What is the reasoning behind passing in parts of the genesis file instead of just passing the entire genesis file?

ebuchman: It is the entire genesis file (ok, except app_hash). Somehow feels better than passing the genesis file as a blob of bytes, no ?

Issue #273, opened 2018-06-16
Switch to list of lists of tags for transaction results, begin block, and end block


cwgoes:
Users will want to query "events" - parameterized logs of particular state transitions, such as account debits/credits, delegations, slashes, etc. - and any number of events may occur per transaction or per begin block / end block function.

melekes:
> I liked the simplicity of a single list of tags, but I can see how groups of tags would be more powerful for advanced queries. Would the indexing code be overhauled for this new system, e.g. querying based on the parameters of an event?

do we need these advanced queries before launch?

what do we need exactly? couldn't it be done with small modifications?

> Besides this test file, I can't find any documentation on the indexing system / query syntax.

https://godoc.org/github.com/tendermint/tendermint/rpc/core#Subscribe
https://github.com/tendermint/tendermint/blob/master/libs/pubsub/query/query.peg

> whether or not particular queries would be efficient

the indexer we're using right now is pretty dumb. if we're talking about fully fledged event system, we will need indexes (BTrees) for tags. In that case, it might make sense to implement https://github.com/tendermint/tendermint/issues/1161, so we don't reinvent the wheel.

zramsay:
https://github.com/tendermint/tendermint/issues/1859

Issue #272, opened 2018-06-15
Add codespace string to error code


cwgoes:
Will remove the necessity to work around error codes with codespacing in the SDK.

edit: Write ADR first.

zramsay:
https://github.com/tendermint/tendermint/issues/1858

PR #279, opened 2018-06-21
Error string instead of codespacing, list of lists of tags

cwgoes:
Closes https://github.com/tendermint/abci/issues/273
Closes https://github.com/tendermint/abci/issues/272

Will also include an ADR but want to update downstream first. Also, maybe the ADR should go in Tendermint since ABCI is being moved there?


cwgoes:
Ready for review. Would be cool if this could be merged before ABCI is moved into Tendermint.

cwgoes:
Closing in favor of https://github.com/tendermint/tendermint/pull/1803.

